PR #56, opened 2019-10-10
Refactor Term documentation to clarify applicative/monadic distinction

CraigFe:
General improvements to doc comments in `lib_term/s.ml`.


CraigFe:
Even clearer would be to scope these inside `Monad` / `Applicative` modules and then explicitly `include` them all.

talex5:
Looks good - thanks!

PR #72, opened 2019-10-22
Add plugin for linting via Dune

CraigFe:
This supplies a linting pipeline stage via the `dune build @fmt` target, which will pre-install the correct version of OCamlformat necessary to run the format:

image

Currently untested.

### DIFF
--- a/plugins/lint/current_lint.ml
+++ b/plugins/lint/current_lint.ml
@@ -1,0 +1,85 @@
+open Current.Syntax
+module Docker = Current_docker.Default
+
+let src = Logs.Src.create "current_fmt" ~doc:"Current pipelines for linting"
+
+module Log = (val Logs.src_log src : Logs.LOG)
+
+let option_map f = function Some x -> Some (f x) | None -> None
+
+let version_from_string =
+  let re =
+    Re.(
+      seq
+        [
+          start;
+          rep space;
+          str "version";
+          rep space;
+          char '=';
+          rep space;
+          group (rep1 @@ diff graph (set "#"));
+          rep space;
+          eol;
+        ]
+      |> compile)
+  in
+  fun path -> Re.exec_opt re path |> option_map (fun g -> Re.Group.get g 1)
+
+let version_from_file path =
+  let ( let+ ) = Lwt.Infix.( >>= ) in
+  let ( let* ) = Lwt.Infix.( >|= ) in
+  if not (Sys.file_exists path) then
+    let () = Logs.info (fun m -> m "No .ocamlformat file found") in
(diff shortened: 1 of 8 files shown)


plugins/lint/current_lint.ml
```diff
@@ -0,0 +1,86 @@
...
+          rep space;
+          eol;
+        ]
+      |> compile)
+  in
+  fun path -> Re.exec_opt re path |> option_map (fun g -> Re.Group.get g 1)
+
+let version_from_file path =
+  let ( let+ ) = Lwt.Infix.( >>= ) in
+  let ( let* ) = Lwt.Infix.( >|= ) in
```
talex5: These are really confusing names! `let*` is supposed to be `>>=` and `let+` should be `>|=` (according to https://github.com/ocaml/ocaml/pull/1947):

> So we use let+ for map, let* for bind and set both the ands to the monoidal product.

plugins/lint/current_lint.ml
```diff
@@ -0,0 +1,86 @@
...
+    match versions with
+    | [ v ] ->
+        let () =
+          Logs.info (fun m -> m "Found OCamlformat version '%s' in dotfile" v)
+        in
+        Ok (Some v)
+    | _ -> Error (`Msg "Unable to parse .ocamlformat file")
+
+let get_ocamlformat_version ~src =
+  Current.component "Infer OCamlformat\nversion"
```
talex5: You should probably use `@,` to force a line-break here. Though it might be better to use `@ ` as a break hint and let it wrap it itself. If it's not wrapping in a sensible place, it probably just needs a change to `component` to set a sensible default wrap width.

plugins/lint/current_lint.ml
```diff
@@ -0,0 +1,86 @@
...
+  |> let> src = src in
+     let switch = Current.Switch.create ~label:"blah" () in
+     let config = Current.Config.v () in
+     let job = Current.Job.create ~switch ~config ~label:"blah" () in
+     Current.Input.const
+       (Lwt_main.run
```
talex5: You can't use `Lwt_main.run` here. OCurrent is already running a mainloop. You'll probably want to use `Current_cache` to handle lwt processes.

CraigFe:
> This functionality seems specific to ocaml-ci, so I'd put it there instead. There's a handy https://github.com/ocaml-ci/ocaml-ci/blob/master/service/analyse.ml module you could add it to...

Indeed; I did look in there at one point and decided against it for reasons I can no longer remember. With fresh eyes, it seems best to put it in there :slightly_smiling_face: Closing here to open in `ocaml-ci``.

PR #78, opened 2019-10-25
Improve diagram for list_map on empty list

talex5:
/cc @CraigFe

### DIFF
--- a/lib_term/analysis.ml
+++ b/lib_term/analysis.ml
@@ -36,7 +36,7 @@ module Make (Job : sig type id end) = struct
   }
   and metadata_ty =
     | Constant of string option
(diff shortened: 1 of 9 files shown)


lib_term/analysis.ml
```diff
@@ -67,7 +67,7 @@ module Make (Job : sig type id end) = struct
     make ~env (Map_input {source; info})
       (match info with
        | Ok _ -> Pass
-       | Error `Blocked -> Blocked)
+       | Error (`Blocked | `Empty_list) -> Blocked)
```
CraigFe: We should be careful that this doesn't produce ambiguous DAGs

talex5: It displays `(empty list)` on the input too, which should make it clear.

PR #79, opened 2019-10-29
Add Current.map_error

talex5:
This is similar to `Result.map_error` in the stdlib.


CraigFe:
We should probably have tests that demonstrate the behaviour of this combinator w.r.t. analysis.

PR #80, opened 2019-10-29
Add Current.option_map

talex5:
On the diagrams, this is shown as a dotted box, indicating that the sub-pipeline can be grey even though the input is ready.

This is a bit simpler than using a list: we don't need to label the items (because there is only even one) and the caller doesn't have to worry about getting multiple items back.

/cc @CraigFe

### DIFF
--- a/lib_term/analysis.ml
+++ b/lib_term/analysis.ml
@@ -37,6 +37,7 @@ module Make (Job : sig type id end) = struct
   and metadata_ty =
     | Constant of string option
     | Map_input of { source : t; info : (string, [`Blocked | `Empty_list]) result }
+    | Opt_input of { source : t; info : [`Blocked | `Selected | `Not_selected] }
     | State of t
     | Catch of t
     | Map_failed of t      (* In [map f t], [f] raised an exception. *)
@@ -45,6 +46,7 @@ module Make (Job : sig type id end) = struct
     | Pair of t * t
     | Gate_on of { ctrl : t; value : t }
     | List_map of { items : t; fn : t }
+    | Option_map of { item : t; fn : t }
 
   type env = {
     next : int ref;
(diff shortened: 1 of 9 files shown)


CraigFe:
The test case doesn't currently expose the behaviour in the `None` case, which I think should probably be tested.

image

image

Also, I think the solution of surrounding the option pipeline with a dashed box is confusing. In particular because the `map` can take on dependencies that will be executed:

image

image

This will happen in the case of the linting step in OCaml-CI, since that requires pulling a Docker image.

I suggest using a dashed arrow to denote a transition that was not taken, and then visually distinguish any deselected blocks from those awaiting the results of prior computations (perhaps using a lighter shade of gray and/or a dotted border).

talex5:
Using dotted arrows (which the previous version of this PR did) is a bit tricky because there are lots of places where we simplify the diagrams by combining arrows. The rule here is that arrows going into a dotted group might not activate their target even though the source is green, and arrows leaving a dotted group can activate their target even though the group wasn't taken. It should be quite easy to colour the group box to indicate whether it was selected or not.

CraigFe:
My issues with the box solution are that arrows in and out have ambiguous behaviour -- they may be direct dependencies / pre-requisites for other steps or not. This seems like quite a brittle approach; what will happen when there are multiple arrows going into such a block? My attempt to produce such an example very quickly deteriorates:

```ocaml
let lint src ~dep ~linter =
  Current.component "lint" |>
  let** _ = src
  and* _ = dep
  and* _ = linter in
  Current.return ()

let v6 commit =
  (* ... *)
  let dep = Current.return ~label:"dep" () in
  let+ _ = Current.option_map (fun linter -> lint src ~dep ~linter) analysis
  and+ _ = Current.option_map (fun linter -> lint src ~dep ~linter) analysis
```
image

Can you give an example in which collapsing of arrows is problematic in this case?

talex5:
It didn't work in the case of your linting PR (https://github.com/ocaml-ci/ocaml-ci/pull/49), because "build OCamlformat" takes its input from a hidden pair node, which groups the arrows from the `pull` and `Analyse`, so they ended up with the same style. That could be fixed with a bit more work, I expect. Feel free to experiment!

The dotted box is currently more of a warning that there's something unusual going on here, but in general the diagrams won't convey all the information that's in the source code.

PR #82, opened 2019-10-30
Autogenerate dune rules for the pipeline expect tests

CraigFe:
- any `test/expected/*.dot` files cause a corresponding dune diff rule to be generated automatically 
 - these diff rules are independently executable, so it's no longer necessary to repeatedly `dune runtest --autopromote` once per changed test-case

Causes `dune runtest --force` not to work (it re-runs the expected output diffs, but not the test-suite itself). I have added a `make test-force` target to recover the old behaviour.

Thanks to @NathanReb and @emillon for their assistance with Dune.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,10 @@ test:
 	dune build @check
 	if dune runtest --no-buffer; then make svgs; else make svgs; exit 1; fi
 
+test-force:
+	rm _build/default/test/*.dot
+	dune runtest --no-buffer -j 1
+
 svg:
 	mkdir svg
 
--- a/test/dune
+++ b/test/dune
@@ -1,37 +1,23 @@
 (executables
-  (names test test_monitor test_cache)
+  (names gen_dune_rules test test_monitor test_cache)
   (libraries current fmt fmt.tty logs.fmt bos lwt.unix alcotest-lwt
              current_fs
              current_docker_test
              current_git_test
              current_opam_test))
 
+(include dune.inc)
+
 (rule
- (targets
-   v1.1.dot v1.2.dot v1.3.dot
-   v1c.1.dot v1c.2.dot v1c.3.dot
-   v2.1.dot v2.2.dot v2.3.dot v2.4.dot
-   v3.1.dot v3.2.dot v3.3.dot
-   v4.1.dot v4.2.dot v4.3.dot
-   v5.1.dot v5.2.dot v5.3.dot
-   v5n.1.dot v5n.2.dot v5n.3.dot
-   option-some.1.dot option-some.2.dot
-   option-none.1.dot option-none.2.dot
- )
- (action  (run ./test.exe)))
+ (targets dune.gen)
+ (deps
+  (source_tree .))
+ (action
+  (with-stdout-to
+   %{targets}
+   (run ./gen_dune_rules.exe))))
 
 (alias
  (name runtest)
- (package current)
  (action
-  (progn
-   (diff expected/v1.1.dot v1.1.dot) (diff expected/v1.2.dot v1.2.dot) (diff expected/v1.3.dot v1.3.dot)
-   (diff expected/v1c.1.dot v1c.1.dot) (diff expected/v1c.2.dot v1c.2.dot) (diff expected/v1c.3.dot v1c.3.dot)
-   (diff expected/v2.1.dot v2.1.dot) (diff expected/v2.2.dot v2.2.dot) (diff expected/v2.3.dot v2.3.dot) (diff expected/v2.4.dot v2.4.dot)
-   (diff expected/v3.1.dot v3.1.dot) (diff expected/v3.2.dot v3.2.dot) (diff expected/v3.3.dot v3.3.dot)
-   (diff expected/v4.1.dot v4.1.dot) (diff expected/v4.2.dot v4.2.dot) (diff expected/v4.3.dot v4.3.dot)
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -6,6 +6,10 @@ test:
 	dune build @check
 	if dune runtest --no-buffer; then make svgs; else make svgs; exit 1; fi
 
+test-force:
+	rm _build/default/test/*.dot
+	dune runtest --no-buffer
```
emillon: no-buffer without j1 will cause interleaving in the output.

test/dune
```diff
@@ -1,37 +1,22 @@
...
-   v5.1.dot v5.2.dot v5.3.dot
-   v5n.1.dot v5n.2.dot v5n.3.dot
-   option-some.1.dot option-some.2.dot
-   option-none.1.dot option-none.2.dot
- )
- (action  (run ./test.exe)))
+ (targets dune.gen)
+ (deps (source_tree .))
+ (action
+  (progn
```
emillon: that progn is unnecessary.

test/gen_dune_rules.ml
```diff
@@ -0,0 +1,30 @@
+let is_testcase name =
+  Fpath.(has_ext "dot" (v name))
+
+let pp_targets ppf filenames =
+  Fmt.pf ppf
+    {|(rule
+ (targets %a)
+ (action (run ./test.exe)))
```
emillon: that part looks a bit weird to me and looks like the core of the problem: `test.ml` is an alcotest test suite, but also outputs dot files that you diff afterwards. Is that correct?

If that's the case, I think it'll be easier to remove this hybrid approach either by:
- using an alcotest assert in `test.ml`. If you want a promote-based workflow, I suppose that using ppx_expect inside `test.ml` would work as well.
- or splitting completely the part that emits dot files from the alcotest test suite (not sure if possible)

CraigFe: Yes, that's correct. There is an Alcotest suite that outputs dotfiles with some expected equivalent, and currently we're using Dune to diff them in order to benefit from a promote workflow.

I don't think it's possible to split the dot file emission from the Alcotest suite. I wasn't aware that it was possible to use `ppx_expect` to benefit from promotion; how would that work?

emillon: The idea is that you write bits like that for every dot file:

```ocaml
let dot = generate_dot "test name" in
print_endline (Dot.to_string dot);
[%expect]
```

And it will diff your ml file against a generated ml file where the output of Dot.to_string has been spliced into the `[%expect]` node. It can then be promoted. (the dot file will be inside a string literal). It's a bit unusual to mix ppx_expect and alcotest like that but should work.

CraigFe: I see; nice. In this case, I think the current solution is probably good enough without taking on an additional test dependency. My only real issue with it is solved by the `test-force` Makefile target. 

Having the expected outputs separate from the test code is nice because it makes it easy to inspect them by e.g. converting a particular file to SVG.

talex5:
CI is failing. Did something change in which package runs the tests?

CraigFe:
I have no idea why this is failing; will dig into it later.

test/gen_dune_rules.ml
emillon: after giving it a bit more thought, combining alcotest and ppx_expect wouldn't work as expectations need to run under a toplevel `let%expect_test ...`.

It seems that what you're after is `Alcotest.check` where the `expected` part is a string literal that's backed by an external file which knows about promotion. Which is not really possible today unfortunately.

CraigFe:
> CI is failing. Did something change in which package runs the tests?

Now fixed by https://github.com/ocaml-ci/ocurrent/pull/82/commits/acb6d673d2f98c5d750bab417e69e9d494398f44.

test/gen_dune_rules.ml
```diff
@@ -0,0 +1,30 @@
...
+
+|}
+    Fmt.(list ~sep:(const string "\n          ") string) filenames
```
talex5: Nit: you can use `@[<v>%a@]` in the format string and let fmt add the indentation (just use `Fmt.cut` as the separator).

test/gen_dune_rules.ml
```diff
@@ -0,0 +1,30 @@
...
+    filename filename
+
+let pp_duneinc = Fmt.(pp_targets ++ (list ~sep:Format.pp_force_newline pp_runtest))
+
+let () =
+  Sys.readdir "expected"
+  |> Array.to_list
+  |> List.sort String.compare
+  |> List.filter is_testcase
+  |> pp_duneinc Fmt.stdout
```
talex5: I suspect you should be using something like `Fmt.pr "@[<v>%a@]@."` here to flush the output at the end.

test/gen_dune_rules.ml
CraigFe: This change breaks left-alignment with the first list item:
```
 (rule
  (targets option-none.1.dot
           ...
-          v5n.1.dot
-          v5n.2.dot
-          v5n.3.dot)
+                option-none.2.dot
+                option-some.1.dot
+                option-some.2.dot
```
Currently trying to work out why...

talex5: Probably due to hard-coding some `\n`s elsewhere - Fmt doesn't like it when you add your own newlines without telling it.

PR #97, opened 2019-11-22
Make transitive dependencies explicit

CraigFe:
This will avoid issues with transitive dependencies changing in `opam-repository`. See https://github.com/ocaml/opam-repository/pull/15334#issuecomment-557265335.

/cc @rgrinberg.


talex5:
Should this be adding `(implicit_transitive_deps false)` to `dune-project` too to check it?

If I do that, it finds additional problems, e.g.

```
dune build @check
File "test/plugins/git/current_git_test.mli", line 7, characters 13-18:
7 |   val pp : t Fmt.t
                 ^^^^^
Error: Unbound module Fmt
```

rgrinberg:
Indeed, I suggest you add this setting to the dune-project file.

CraigFe:
Quite. I was convinced I had done that... Good spot.

PR #99, opened 2019-12-17
Fetch logs and files from dead containers

Julow:
Hi !

This is a draft PR trying to fetch results from dead containers after they ran.

With the current implementation, the whole files are read in memory and put in the cache.

The `Container_result` abstraction is only implemented for the interface and is not necessary.

Example:
```ocaml
  let+ file1, (file2, logs) =
    Docker.run image ~args ~results:Container_result.(file path1 & file path2 & logs ())
  in
  ...
```

It's currently untested.

### Alternative

My first idea was to pass a function to `run`:
```ocaml
val run : Image.t -> ... -> post_run:(container_id -> status:int -> 'a Current.t) -> 'a Current.t
```
`container_id` (abstract, should not escape) allows to call docker commands on the dead container (eg. log, cp, commit).
This way, only the data the user need is cached instead of whole files.
The problem of this approach is that the user have to pass marshal/unmarshal functions for his `'a` and changing the function won't invalidate the cache.

### DIFF
--- a/plugins/docker/current_docker.ml
+++ b/plugins/docker/current_docker.ml
@@ -7,6 +7,51 @@ let pp_tag = Fmt.using (Astring.String.cuts ~sep:":") Fmt.(list ~sep:(unit ":@,"
 module Make (Host : S.HOST) = struct
   module Image = Image
 
+  module Container_result = struct
+
+    type 'a t =
+      | Logs : Run.fetch_logs -> string list t
+      | File : { path : Fpath.t } -> string t
+      | Unit : unit t
+      | Pair : 'a t * 'b t -> ('a * 'b) t
+
+    let (&) a b = Pair (a, b)
+    let unit = Unit
+    let logs ?tail () = Logs { tail }
+    let file path = File { path }
+
+    let rec to_yojson : type a. a t -> _ = function
+      | Logs { tail } ->
+        let tail = match tail with Some t -> `Int t | None -> `String "none" in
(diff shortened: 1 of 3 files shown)


kit-ty-kate:
I'm encountering the same limitations of not being able to get the result of `Docker.run` in `opam-ci`.

Looking at the PR itself, wouldn't it be simpler to just pipe the result directly (to either a file or a string) instead of calling `docker logs`? That would reduce the need for docker to store even more data than necessary and would avoid dealing too much with docker.

Julow:
The idea of this PR was to encapsulate interactions with docker but still give access to every "results" (logs, files in the container's FS).

Docker doesn't store more data, the dead container is removed after we got what we need (as before).
Except for the failed attempt at being smart (`Container_result`), this PR is as simple as calling `docker logs` or `docker cp`.

talex5:
I'm worried that this seems a bit ad-hoc. As @CraigFe pointed out, it's likely we'll want other things (such as streaming). I've just merged #130, which should make it easier to add custom pipeline stages that use Docker.

Here's an example showing how to use it. It runs nginx in a container, and then (while it's running) execs a curl process also inside the container to test it. I think that should be flexible enough to cover your cases too.

https://github.com/ocurrent/ocurrent/blob/9aa4f4fba6de52d54ec39a3985a584c5117ca4e4/examples/docker_custom.ml#L17-L55

Julow:
I agree new API is more powerful. At the expense of adding (complex) boilerplate.

We should close this PR.

PR #104, opened 2020-01-28
Github plugin: Expose the origin (branch or PR) of a commit

kit-ty-kate:
Using the github plugin, it is currently rather hard to know whether a commit is from a PRs or a Branch as both are merged together and their kind are not exposed to the user level.

Comparing their `Repo_id.t` with the main repository might have been a solution, however this still does not differentiate between a branch and a PR where the branch used to open it, is from the repository itself.

This PR adds the `kind` function which exposes the internal `id` field and allows users to have some more informations about the origin of a commit

### DIFF
--- a/plugins/github/api.ml
+++ b/plugins/github/api.ml
@@ -512,6 +512,8 @@ module Commit = struct
 
   let hash (_, id) = id.Commit_id.hash
 
+  let kind (_, id) = id.Commit_id.id
+
   let pp = Fmt.using snd Commit_id.pp
 
   let set_status commit context status =
--- a/plugins/github/api.mli
+++ b/plugins/github/api.mli
@@ -13,6 +13,7 @@ module Commit : sig
   val repo_id : t -> Repo_id.t
   val owner_name : t -> string
   val hash : t -> string
+  val kind : t -> [ `Ref of string | `PR of int ]
   val pp : t Fmt.t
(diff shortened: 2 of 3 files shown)


talex5:
I suspect there is a bit of confusion in the original API here. In general, a single commit may belong to multiple branches and PRs. However, as GitHub doesn't provide a way to fetch a commit by ID, we store one of the branches or PRs that contain it so we can fetch it if needed.

So a `Commit.t` is really a commit combined with a ref. But `Commit.equal` just compares the hash. This works mostly OK if you only use the information for fetching, but it's rather confusing. Maybe we should rename `Commit` to `Ref_commit` or something and change `equal` to consider the same commit on separate refs to be distinct.

Also, note that you can only set the status of a commit, not a branch or PR. In particular, you cannot perform one set of tests on PRs and a different set of tests on branches and expect to set the same commit status key (one will overwrite the other if a branch and PR have the same head).

kit-ty-kate:
Is there anything I can do to improve this PR? I'm not sure how to proceed.

talex5:
The code has changed a bit now. Internally, it keeps map from ref names to commits, so it should be pretty easy to add a function that does what you want:

https://github.com/ocurrent/ocurrent/blob/63e92bcd23ebe42b6f01315366863cfcb64f30b2/plugins/github/api.ml#L413-L425

talex5:
Closing in favour of #123.

PR #105, opened 2020-01-28
Github API: Split ci_refs into Branches and PRs

kit-ty-kate:
This is a draft PR that stores branches and prs independently. Currently both are stored together and delivered via the the `ci_refs` function. However, if feel like this function tries to do to many things and its name does not clearly reflect what it does.

My proposition is to split this function into two functions exposed to the user `branches` and `prs` which returns every branches from a given repo or every PR from a given repo.

However one annoying side effect of this split is code duplication and query duplication in case both branches and PRs are required. However this reduces the size of the request in case only one is required.

This is only a draft as an invitation to discuss what can be done to improve the github API. The main feature I personally would like to have is in a separate PR (https://github.com/ocurrent/ocurrent/pull/104) so this is not urgent.

### DIFF
--- a/plugins/github/api.ml
+++ b/plugins/github/api.ml
@@ -121,16 +121,18 @@ type t = {
   token_lock : Lwt_mutex.t;
   mutable token : token;
   mutable head_inputs : commit Current.Input.t Repo_map.t;
-  mutable ci_refs_inputs : ci_refs Current.Input.t Repo_map.t;
+  mutable branches_inputs : ci_refs Current.Input.t Repo_map.t;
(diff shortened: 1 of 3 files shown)


talex5:
I don't remember the details, but I think that doing one call may be cheaper in terms of GitHub app points (which limits how many queries you can do). In general, the idea of GraphQL is to get all the information in a single request. It might be worth exposing more control over what goes in the query, but it also becomes harder to compose pipeline stages if they each want different subsets of the information, so grabbing it all may be simpler.

talex5:
Can we close this now that #123 is in?

PR #110, opened 2020-02-13
Require Alcotest < 1.0.0

talex5:
The new version doesn't work (will be fixed in 1.0.1).


talex5:
Build passed, but GitHub doesn't seem to want to show the new status (the set-status operation passed though, and I retried it too).

PR #111, opened 2020-02-17
Allow alternative Dockerfile filenames in `Docker.build`

talex5:
Previously, you could only build `Dockerfile`, or specify the contents manually.

### DIFF
--- a/plugins/docker/build.ml
+++ b/plugins/docker/build.ml
@@ -17,15 +17,15 @@ let use_pool pool f =
 module Key = struct
   type t = {
     commit : [ `No_context | `Git of Current_git.Commit.t ];
-    dockerfile : string option;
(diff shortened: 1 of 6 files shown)


plugins/docker/build.ml
```diff
@@ -17,15 +17,15 @@ let use_pool pool f =
 module Key = struct
   type t = {
     commit : [ `No_context | `Git of Current_git.Commit.t ];
-    dockerfile : string option;
+    dockerfile : [`File of string | `Contents of string];
```
CraigFe:
```suggestion
    dockerfile : [`File of Fpath.t | `Contents of string];
```
Is there a reason not to use `Fpath.t` here? It seems cleaner to me.

talex5: Good idea; I've updated it.

PR #118, opened 2020-02-27
Report a Prometheus metric for metrics collection time

talex5:


talex5:
I'm not sure whether this will be useful in future. I'm testing https://github.com/mirage/prometheus/pull/25 with it, and if that fixes the collection slowness then we'll probably never need this metric again.

talex5:
Updating to the new version of Prometheus has fixed this:

fixed

PR #120, opened 2020-03-03
Make Docker.run return its stdout as a string

kit-ty-kate:
After commenting on https://github.com/ocurrent/ocurrent/pull/99#issuecomment-593637717 I realized it was relatively easy to change `Current_docker.run` to do what i want and return its result as a string.

This is the current design I'm using in opam-ci where it successfully gets the list of reverse dependencies for each packages:
screenshot

### DIFF
--- a/plugins/docker/dune
+++ b/plugins/docker/dune
@@ -16,5 +16,5 @@
    ppx_deriving_yojson.runtime
    result)
  (preprocess (per_module
-              ((pps ppx_deriving.std ppx_deriving_yojson) pull build run tag push service)
+              ((pps ppx_deriving.std ppx_deriving_yojson) pull build run pread tag push service)
 )))
--- a/plugins/docker/pread.ml
+++ b/plugins/docker/pread.ml
@@ -1,0 +1,41 @@
+open Lwt.Infix
+
+type t = {
+  pool : Current.Pool.t option;
+}
+
+let id = "docker-pread"
+
+module Key = struct
+  type t = {
+    image : Image.t;
+    args : string list;
+    docker_context : string option;
+    run_args : string list;
+  }
+
+  let pp_args = Fmt.(list ~sep:sp (quote string))
+
+  let cmd { image; args; docker_context; run_args } =
+    Cmd.docker ~docker_context @@ ["run"] @ run_args @ ["--rm"; "-i"; Image.hash image] @ args
+
+  let pp f t = Cmd.pp f (cmd t)
+
+  let digest { image; args; docker_context; run_args } =
+    Yojson.Safe.to_string @@ `Assoc [
+      "image", `String (Image.hash image);
+      "args", [%derive.to_yojson:string list] args;
+      "docker_context", [%derive.to_yojson:string option] docker_context;
+      "run_args", [%derive.to_yojson:string list] run_args;
+    ]
+end
+
+module Value = Current.String
+
+let build { pool } job key =
+  Current.Job.start job ?pool ~level:Current.Level.Average >>= fun () ->
+  Current.Process.check_output ~cancellable:true ~job (Key.cmd key)
+
+let pp = Key.pp
+
+let auto_cancel = true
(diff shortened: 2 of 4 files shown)


talex5:
This looks good, but I think it should be a new function (e.g. `Docker.pread`), as we often don't need to use (or store) the output.

CraigFe:
Thanks :slightly_smiling_face: Some thoughts:
 - the use of `Current.Process.check_output` rather than `docker logs` sacrifices the ability to tail the logs, which is perhaps significant for very containers with very verbose output. I'm not sure how `docker logs --tail` is implemented, but it seems likely to be better than loading the entire log as a string and then truncating it.

- `Current.Process.check_output` seems to _redirect_ the `stdout` stream to a string rather than the Current logs, which is problematic as it means those logs are no longer visible in the UI.

 - `Current.Process.check_output` does not capture the `stderr` stream, unlike the previous attempt at a `docker logs` solution.  right? If so, it should probably be added to the

talex5:
@CraigFe I think this use is quite separate to log analysis. For example, Kate wants to read structured data (revdeps) from the command. This should not be mixed in with warning messages from stderr. It's also quite likely that the output being captured would be binary data, in which case it's important that we *don't* log it.

CraigFe:
@talex5 : The use-cases I can think of would all want the data to be logged as well as captured. This includes the application to benchmarking (cc @gs0510, @MagnusS), who have been waiting for a feature like this for a while. Is it an issue if the rev-deps, in this case, are _not_ logged to the interface? I think as a new user of OCurrent I would expect them to be.

If we have a use-case for extracting binary data from a Docker container, I suggest doing that by some other mechanism than overloading 'logs'. For instance by using `docker cp` (as was attempted in the other PR), or volume mounting.

plugins/docker/pread.ml
```diff
@@ -0,0 +1,41 @@
+open Lwt.Infix
+
+type t = {
+  pool : Current.Pool.t option;
+}
+
+let id = "docker-run"
```
talex5: Needs a different ID here.
```suggestion
let id = "docker-pread"
```

plugins/docker/dune
```diff
@@ -16,5 +16,5 @@
    ppx_deriving_yojson.runtime
    result)
  (preprocess (per_module
-              ((pps ppx_deriving.std ppx_deriving_yojson) pull build run tag push service)
+              ((pps ppx_deriving.std ppx_deriving_yojson) pull build run pread tag push service)
```
talex5: (we could probably get rid of this per-module thing now - it was a hack for before 4.08, which we no longer support)

kit-ty-kate: Should I change this in this PR or in a separate PR?

kit-ty-kate:
If the change in `plugins/docker/dune` can be done separately, this should be good to go once CI is green.

talex5:
Thanks - I'll take a look at cleaning up the preprocessing separately...

CraigFe:
I don't think this should have been merged.

- The logging behaviour may be problematic for our other use-case, and at the very least seems to warrant further discussion.
- `plugins/docker/pread.ml` duplicates all of the code from `plugins/docker/run.ml`, when this seems unnecessary (they share the same `Key` module, for instance). 
- It's missing a test in `test/plugins/docker/`, ideally capturing the behaviours w.r.t. logging and `stderr`.
- The docstring for `pread` doesn't explain that the logging behaviour is not the same as `run`.

Are we in a rush to merge this feature as is?

talex5:
The behaviour of pread (returning the output instead of sending it to stdout) is pretty standard, e.g. compare lwt's version:

```ocaml
utop # #require "lwt.unix";;
utop # open Lwt_process;;
utop # let cmd = "", [| "echo"; "hello" |];;

utop # exec cmd;;
hello
- : Unix.process_status = Unix.WEXITED 0

utop # pread cmd;;
- : string = "hello\n"
```

But I do think we should make it easier to make user-defined pipeline stages using Docker, so the built-in ones don't have to cover every case.

PR #121, opened 2020-03-04
Update to new Alcotest API

talex5:


test/test.ml
```diff
@@ -187,24 +187,28 @@ let test_all_labelled () =
...
-    "terms", [
-      Alcotest.test_case "all_labelled" `Quick test_all_labelled;
-    ];
-    "cache", Test_cache.tests;
-    "monitor", Test_monitor.tests;
-    "job", Test_job.tests;
-    "log_matcher", Test_log_matcher.tests;
-  ]
+  Lwt_main.run begin
+    Cli.run "test" [
```
samoht: Hum that looks more complicated that it should. @CraigFe is there any simpler way to do this?

talex5: I don't think there is, because nothing else in Alcotest refers to Cli except for `alcotest.ml:module T = Cli.Make (Monad.Identity)`, which isn't using `Lwt`.

CraigFe: Ideally it would be `Alcotest_lwt.run` rather than constructing `Cli.run`, but unfortunately it doesn't work right now because of a bad revert: https://github.com/mirage/alcotest/pull/182#issuecomment-530031614 should have done something similar for the concurrent backends (instantiating with `Cli` rather than `Core`). I'll make a PR and a patch fix in Alcotest.

PR #122, opened 2020-03-04
Add Prometheus metric for in-memory cache sizes

talex5:
The fact that we only ever add things to the cache is likely to be a problem, but let's add monitoring first before trying to fix it...


talex5:
(included in #125)

PR #123, opened 2020-03-04
Github: expose the low-level refs function

talex5:
This allows users to make custom pipeline stages similar to `ci_refs` and `head_of`.

@kit-ty-kate : would this solve your problem in #104?

### DIFF
--- a/plugins/github/api.ml
+++ b/plugins/github/api.ml
@@ -89,7 +89,7 @@ let no_token = {
   expiry = Some (-1.0);
 }
 
-module GH_ref = struct
+module Ref = struct
   type t = [ `Ref of string | `PR of int ] [@@deriving to_yojson]
 
   let compare = Stdlib.compare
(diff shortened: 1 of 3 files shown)


kit-ty-kate:
It solves my problem but I'm not sure it's the right solution/interface. Compared to my code with #104, with this PR, the code seems rather complicated and I spent a good chunk of time to get my head around the `Current.Input.t` type.
Here the diff of my code compared with #104:
```diff
-let get_prs repo =
-  let* refs = Github.Api.Repo.ci_refs repo in
-  List.fold_left begin fun acc head ->
-    let+ acc = acc in
-    match Github.Api.Commit.kind head with
+let get_prs installation repo =
+  let* installation = installation
+  and* repo = repo in
+  let api = Github.Installation.api installation in
+  let repo = Github.Api.Repo.id repo in
+  let+ refs =
+    Current.component "Github API refs" |>
+    let> () = Current.return () in
+    Github.Api.refs api repo
+  in
+  Github.Api.Ref_map.fold begin fun key head acc ->
+    match key with
     | `Ref _ -> acc (* Skip branches, only check PRs *)
     | `PR _ -> head :: acc
-  end (Current.return []) refs
+  end refs []
```

kit-ty-kate:
Also for users who want to do the other way around (getting only branches), it requires them to again filter out the `gh-page` branch manually

talex5:
Well, the problem with #104 is that, logically, a Git commit doesn't have a unique "branch", so it doesn't make sense to provide an API that returns one, even though for hacky internal reasons it might work in your case. This API reverses the mapping: it says that each branch has exactly one head commit, rather than saying that each commit has exactly one branch.

BTW, the reason for exposing it as an input is partly so you can provide your own label for it (e.g. "PRs"), and also so you can decide which of the inputs should be dynamic. We can add other helpers around this if they turn out to be useful, but this is the lowest-level API.

kit-ty-kate:
I just realized my code can be simplified a whole lot and the diff is much more reasonable now.
```diff
 let get_prs repo =
-  let+ refs = Github.Api.Repo.ci_refs repo in
-  List.fold_left begin fun acc head ->
-    match Github.Api.Commit.kind head with
+  let+ refs =
+    Current.component "Get PRs" |>
+    let> (api, repo) = repo in
+    Github.Api.refs api repo
+  in
+  Github.Api.Ref_map.fold begin fun key head acc ->
+    match key with
     | `Ref _ -> acc (* Skip branches, only check PRs *)
     | `PR _ -> head :: acc
-  end [] refs
+  end refs []
```
My comment about filtering-out `gh-page` for people who want the list of branches only still stands but this can be improved later I think. This PR looks good as is :+1:

PR #125, opened 2020-03-04
Report information about active jobs

talex5:
- Add a simple /jobs page to the admin web UI
- Report a Prometheus metric for number of active jobs


talex5:
Neither of these metrics explains the growth in heap size, but worth having anyway I think:

metrics

(these graphs show the effect of updating one of the Docker base images in ocaml-ci)

PR #127, opened 2020-03-05
Testing: Add /heap page to admin UI

talex5:
Shows a random sample of heap blocks using the new `Memprof` feature that was (almost) added to OCaml 4.10. The public API was removed as the feature wasn't quite ready, but it might still provide some useful information.

### DIFF
--- a/lib_web/current_web.ml
+++ b/lib_web/current_web.ml
@@ -133,6 +133,8 @@ let handle_request ~engine ~webhooks _conn request body =
               set_confirm (Current.Engine.config engine) data
             | ["log-rules"] ->
               Log_rules.handle_post data
+            | ["heap"] ->
+              Heap.handle_post data
             | _ ->
               Server.respond_not_found ()
           end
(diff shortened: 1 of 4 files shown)


talex5:
This seems pretty useful! After updating one of the Docker base images in CI, the heap mostly consists of:

1. Dockerfiles.
2. `Lwt_unix.sleep` calls.

ocaml-ci tries to avoid doing too much Dockerfile generation by caching the `Dockerfile.t`. Unfortunately, we still need to turn this into a string and then digest it on every evaluation. This isn't a memory leak, but it does involve a `String.concat`, which is slow and generates a lot of short-lived garbage on the heap.

The sleep calls are for the `schedule` option. The main problem here is that `Current_cache`'s `resolved` function generates a new `Input` (and, therefore, a new sleep) on every evaluation. We should cache these. We should probably also cancel the sleep when the input is no longer needed.

talex5:
The memory profiler API has changed in 4.11. Will update to that and reopen the PR if we need it again.

PR #144, opened 2020-03-20
Evaluate pipelines incrementally

talex5:
The first commit adds a small incremental library (a few hundred lines of code, including comments, and no dependencies). It includes some code based on @let-def's orderme library, so I licensed that bit as ISC.

The second commit updates `lib_term` to use this for evaluation, and changes the type for inputs to be incremental too.

I need to do more testing, but it seems to be working so far! Once this is in, there is lots of scope for further improvements (e.g. `list_map` should preserve the existing pipelines where possible), but let's see what this does on its own first.

### DIFF
--- a/current.opam
+++ b/current.opam
@@ -23,6 +23,7 @@ build: [
 ]
 depends: [
   "ocaml" {>= "4.08.0"}
+  "current_incr" {= version}
   "fmt"
   "bos"
   "ppx_deriving"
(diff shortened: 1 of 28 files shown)


let-def:
The code taken from orderme can be relicensed to Apache 2 to match the existing license.

talex5:
Thanks!

PR #145, opened 2020-03-22
Update README.md to link the API Docs

shonfeder:
Just suggesting a slight change to the README to 

- make the documentation very easy to find, by adding a heading
- make license an h2 sub-heading (assuming the h1 **Ocurrent** is supposed to be the title)
- Link the API docs directly from the readme (so users don't have to go hunting for  it through the wiki)


talex5:
Yeah, that looks better - thanks!

PR #167, opened 2020-04-06
Add a context to maps

kit-ty-kate:
Currently, maps – contrary to binds – do not carry a context information. This could be useful in several cases. For instance, see the following sketched code:
```OCaml
let x =
  let+ list_file = Current_docker.pread image "ls distro.txt" in
  let dockerfile = from "%s" distro @@ ... in
  ("example", Current_docker.build ~dockerfile image)
```
Other functions might also benefit from a similar change

### DIFF
--- a/lib_term/current_term.ml
+++ b/lib_term/current_term.ml
@@ -71,11 +71,15 @@ module Make (Metadata : sig type t end) = struct
 
   let map f x =
     let id = Id.mint () in
-    node ~id (Map (Term x)) @@ Current_incr.map (Dyn.map ~id f) x.v
+    node ~id (Map (Term x)) @@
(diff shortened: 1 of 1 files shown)


talex5:
You can't use `x` as the bind context here because it will lose any existing context. You'd have to create some new term (like `bind` does with `Bind_in`) and use that.

But I suspect this isn't very useful, because we want to avoid people using `bind` as much as possible and encourage them to use other combinators (which can be statically analysed). A `map` that produces new pipeline components would need to go through a `bind` at some point to do anything with them, so it's really just a bind in disguise.

What are you trying to do with `Current.map`? Can `Current.list_map` help?

talex5:
I guess this isn't needed now. Please reopen if it is.

PR #186, opened 2020-04-30
port current_github to use mirage-crypto from nocrypto

avsm:


avsm:
Sorry, accidentally from an ancient branch!

PR #190, opened 2020-05-09
use `Routes.empty` for empty routes.

anuragsoni:
Starting version 0.8.0, nil can only be used after another route pattern.

See: https://github.com/anuragsoni/routes/pull/111


anuragsoni:
0.8.0 is available on opam now

PR #192, opened 2020-05-11
Delete directories with Lwt

talex5:
Maybe this will reduce the latency spikes?


talex5:
Well, that helped more than I was expecting!

latency

PR #193, opened 2020-05-13
GitHub: get all installation repositories, not just the first 30

### LINKED ISSUE #139: GitHub plugin is limited to 30 repos
The GitHub API call for getting an installation's repositories only returns one page at a time (seems to be about 30 items). We should fetch the remaining pages. But it would also be good to have some limits in place, in case someone enables "all repositories" by mistake...

talex5:
For now, as a precaution, ignore any installation that has configured all repositories to be tested (this is likely an accident).

Fixes #139.


avsm:
*coughs quietly and stops activating the whole mirage org again*

PR #198, opened 2020-05-17
GitHub: if listing repositories fails, retry after 30s

talex5:
We retry once, using the same token. This is to narrow down where the problem is.


talex5:
It might be a while before this error shows up again and we can test this, so merging it now.

PR #210, opened 2020-07-09
rename "whitelist" to "allowlist"

avsm:
corresponding PR to ocaml-ci incoming once this is merged

### DIFF
--- a/plugins/github/app.ml
+++ b/plugins/github/app.ml
@@ -34,7 +34,7 @@ module Installs = Current.Var(struct
 type t = {
   app_id : string;
   key : Mirage_crypto_pk.Rsa.priv;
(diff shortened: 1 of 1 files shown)


plugins/github/app.ml
```diff
@@ -168,13 +168,13 @@ let app_id =
     ~docv:"ID"
     ["github-app-id"]
 
-let whitelist =
+let allowlist =
   Arg.required @@
   Arg.opt Arg.(some (list string)) None @@
   Arg.info
     ~doc:"A comma-separated list of whilelisted GitHub accounts"
```
SquidDev, line 175: The documentation here should probably also be updated.

avsm: Well spotted, thanks!

PR #211, opened 2020-07-10
Case-insensitive allowlist

Julow:
Draft PR until https://github.com/ocurrent/ocurrent/pull/210 is merged and to rename here too.

Github's API is case insensitive w.r.t usernames so collisions cannot happen (for example: https://api.github.com/users/OCURRENT) and it is easy to use the wrong case when adding someone. (happened to me)

This shouldn't allow any edge cases because no URLs/usernames are created from elements of the whitelist, only values returned by the API are used.


avsm:
#210 is merged, so if you rebase/rename this should be good to go @jullow

talex5:
Thanks!

PR #213, opened 2020-07-14
docker: add an arch tag to the pull command

avsm:
This parses the output of `docker manifest inspect` and picks the appropriate hash from the multiarch manifest, and then pulls that specific tag by its hash.

If not specified, docker pull operates as normal and resolves to the default architecture of the build host.

### DIFF
--- a/plugins/docker/pull.ml
+++ b/plugins/docker/pull.ml
@@ -2,6 +2,8 @@ open Lwt.Infix
 
 type t = No_context
 
+let ( >>!= ) = Lwt_result.bind
+
 module Key = struct
   type t = {
     docker_context : string option;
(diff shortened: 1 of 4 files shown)


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "ocaml-version"]
```
talex5, line 1: Can this be removed now? Looks like 2.6.0 is released.

current_docker.opam
```diff
@@ -24,6 +24,7 @@ depends: [
   "ppx_deriving"
   "lwt"
   "dockerfile"
+  "ocaml-version" {>="2.5.0"}
```
talex5, line 27: Should this be 2.6.0?
```suggestion
  "ocaml-version" {>="2.6.0"}
```

plugins/docker/pull.ml
```diff
@@ -17,19 +32,47 @@ module Value = Image
 
 let id = "docker-pull"
 
+let get_digest_from_manifest manifest arch =
+  let jstring = function `String s -> s |_ ->raise Not_found in
+  let jlist = function `List l -> l |_->raise Not_found in
+  let jassoc = function `Assoc x -> x | _->raise Not_found in
```
talex5, line 38: There are convenience functions for these in yojson, e.g. `Yojson.Basic.Util.to_string`

plugins/docker/pull.ml
```diff
@@ -17,19 +32,47 @@ module Value = Image
...
+  try Yojson.Basic.from_string manifest |>
+  jassoc |> List.assoc "manifests" |> jlist |> List.map jassoc |>
+  List.find (fun l -> List.assoc "platform" l |> jassoc |> fun l ->
```
talex5, line 41: `Yojson.Basic.Util.member`

avsm:
Yes, agreed. I'll just make it take an architecture string and the other changes you suggest.

avsm:
Review comments fixed and ocurrent/ocaml-ci#219 updated to go with it

talex5:
Pushed a couple of changes to re-indent it, improve the error message, and add the argument to the non-raw API too.

PR #218, opened 2020-08-05
Update Dockerfile to latest opam-repository

talex5:
Perhaps we should remove the `Dockerfile`, but it's useful for some of the examples.


avsm:
I've got some code lying around to autogenerate the Dockerfile, will tidy it up later

PR #225, opened 2020-10-20
current_web depends on conf-graphviz

dra27:
I wondered why I wasn't getting pipeline diagrams. I switched browser. I still wasn't pipeline diagrams. I realised why. i felt silly.

Hopefully this is the correct to address any further silly people...

### DIFF
--- a/current_web.opam
+++ b/current_web.opam
@@ -33,4 +33,5 @@ depends: [
   "tyxml"
   "routes" {>= "0.8.0"}
   "dune" {>= "2.0"}
+  "conf-graphviz"
 ]


talex5:
Ah, good idea!

The opam change looks fine, but there's no ocamlfind library called `conf_graphviz` - I guess we can just remove that bit?

dra27:
Oops - I thought opam packages were valid there too! Re-pushed with just the opam file

talex5:
Looks like conf-graphviz doesn't work on opensuse. I'll make a fix...

talex5 (approved):

PR #227, opened 2020-10-24
Allow pushing manifests to insecure Docker registries

dra27:
At the moment, this simply adds `--insecure` to the docker manifest calls. I expect it'd be better to propagate this from higher up, but I wasn't sure exactly where to pass the instruction (bolt it into the `config`, add `?insecure`, etc.?) so it can start as a draft PR!

### DIFF
--- a/plugins/docker/push_manifest.ml
+++ b/plugins/docker/push_manifest.ml
@@ -34,10 +34,10 @@ module Outcome = struct
 end
 
 let create_cmd ~config ~tag {Value.manifests} =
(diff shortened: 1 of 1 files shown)


talex5:
Yeah, it's not obvious. I guess an `?insecure` argument would be best. `auth` is about the user, so it doesn't seem right to mark that as "insecure".

shonfeder:
Just cleaning up some stale PRs. LMK if some one feels we need to land this for some reason.

PR #228, opened 2020-11-03
add Docker Compose support to the Docker plugin

avsm:


plugins/docker/s.ml
```diff
@@ -83,6 +83,11 @@ module type DOCKER = sig
 
   val service : name:string -> image:Image.t Current.t -> unit -> unit Current.t
   (** [service ~name ~image ()] keeps a Docker SwarmKit service up-to-date. *)
+
+  val compose : name:string -> image:Image.t Current.t -> contents:string -> unit -> unit Current.t
+  (** [service ~name ~image ~contents ()] keeps a Docker Compose deployment up-to-date.
+      [contents] contains the full Compose Yaml file, and the [$IMAGE_HASH] variable will be
+       substituted at deployment time with the SHA256 hash to pull. *)
```
talex5, line 90: Probably we should just have `contents:string Current.t` here and let the caller to whatever regex stuff they want. Otherwise, this is limited to composing a single service.

talex5 (commented):
How does docker-compose keep track of state? Do we need to persist it somewhere, so it knows what services are running across restarts?

It seems a shame to have to add this in, instead of using the newer support for Docker stacks, but it seems that stacks don't allow binding different services to different IP addresses.

I wonder if this is a limitation of docker services in general, or just of docker stack?

plugins/docker/s.ml
avsm: Done

plugins/docker/compose.ml
```diff
@@ -0,0 +1,45 @@
...
+
+  let digest { contents } =
+    Yojson.Safe.to_string @@ `Assoc [
+      "contents", `String contents
+    ]
+end
+
+module Outcome = Current.Unit
+
+let cmd args { Key.docker_context; name; _ } =
```
talex5, line 29:
```suggestion
let cmd args { Key.docker_context; name } =
```

talex5 (approved):

PR #231, opened 2020-11-10
Staleness argument

patricoferris:
This PR adds an optional `staleness` argument to `ci_refs` (defaulting to `None`) which allows a user to specify how old, as a number of days, is too old for the last commit of a PR/branch to be considered for using. It was tested locally with a modified `example/github.ml` pipeline. 

It does add `ptime` as a dependency, if this isn't desirable it's possible to write one using just functions on strings?

### DIFF
--- a/plugins/github/api.ml
+++ b/plugins/github/api.ml
@@ -166,6 +166,10 @@ and ci_refs = {
   all_refs: commit Ref_map.t
 }
 
+let get_default_ref t = t.default_ref 
+
+let get_all_refs t = t.all_refs
+
 let v ~get_token account =
   let head_monitors = Repo_map.empty in
   let ci_refs_monitors = Repo_map.empty in
--- a/plugins/github/api.ml
+++ b/plugins/github/api.ml
@@ -476,46 +476,40 @@ let refs t repo =
       i
   )
 
-let to_ci_refs refs =
-  refs
-  |> Ref_map.remove (`Ref "refs/heads/gh-pages")
-  |> Ref_map.bindings
-  |> List.map snd
-
-let now () =
-  Ptime.of_float_s (Unix.gettimeofday ()) |> function
-  | Some t -> t
-  | None -> Fmt.failwith "Failed to get current time"
-
 let to_ptime str =
   Ptime.of_rfc3339 str |> function
   | Ok (t, _, _) -> t
   | Error (`RFC3339 (_, e)) -> Fmt.failwith "%a" Ptime.pp_rfc3339_error e
 
-(** Check if the elapsed time from timestamps [start, finish] is within the [cutoff] duration *)
-let active_date_cutoff ~start ~finish cutoff =
-  let diff = Ptime.diff finish start in
-  match Ptime.Span.to_int_s diff with
-    | Some s -> (Duration.to_sec cutoff) - s > 0
-    | None -> Fmt.failwith "Failed to calculate ptime diff: %a" Ptime.Span.pp diff
+let remove_stale ?staleness ~default_ref refs =
+  match staleness with
+  | None -> refs
+  | Some staleness ->
+    let cutoff = Unix.gettimeofday () -. Duration.to_f staleness in
+    let active x =
+      let committed = Ptime.to_float_s (to_ptime x.Commit_id.committed_date) in
+      committed > cutoff
+    in
+    let is_default = function
+      | { Commit_id.id = `Ref t; _ } -> String.equal default_ref t
+      | _ -> false
+    in
+    List.filter (fun (_, x) -> is_default x || active x) refs
 
-let ci_refs ?(staleness=None) t repo =
+let to_ci_refs ?staleness refs =
+  refs.all_refs
+  |> Ref_map.remove (`Ref "refs/heads/gh-pages")
+  |> Ref_map.bindings
+  |> List.map snd
+  |> remove_stale ?staleness ~default_ref:refs.default_ref
+
+let ci_refs ?staleness t repo =
   let+ refs =
     Current.component "%a CI refs" Repo_id.pp repo |>
     let> () = Current.return () in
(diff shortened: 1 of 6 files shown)


plugins/github/api.ml
```diff
@@ -451,13 +470,36 @@ let to_ci_refs refs =
...
-let ci_refs t repo =
+let now () =
+  Ptime.of_float_s (Unix.gettimeofday ()) |> function
+  | Some t -> t
+  | None -> Fmt.failwith "Failed to get current time"
+
+let to_ptime str =
+  Ptime.of_rfc3339 str |> function
+  | Ok (t, _, _) -> t
+  | Error _ -> Fmt.failwith "Datetime parsing error for: %s" str
```
talex5, line 481: Any reason not to use `Ptime.pp_rfc3339_error` to show the original error here?

plugins/github/api.ml
```diff
@@ -451,13 +470,36 @@ let to_ci_refs refs =
...
+let ci_refs ?(staleness=None) t repo =
   let+ refs =
     Current.component "%a CI refs" Repo_id.pp repo |>
     let> () = Current.return () in
     refs t repo
   in
-  to_ci_refs refs
+  match staleness with 
+    | None -> to_ci_refs refs 
+    | Some n when n < 0 -> to_ci_refs refs
```
talex5, line 497: Does `n < 0` need to be a special case? I think we can just remove this line.

plugins/github/api.ml
```diff
@@ -628,7 +672,7 @@ module Anonymous = struct
       Lwt.try_bind
         (fun () -> query_head repo gref)
         (fun hash ->
-           let id = { Commit_id.owner_name; hash; id = gref } in
+           let id = { Commit_id.owner_name; hash; id = gref; committed_date = "" } in
```
talex5, line 675: Unfinished change here?

plugins/github/api.mli
```diff
@@ -34,7 +35,7 @@ val exec_graphql : ?variables:(string * Yojson.Safe.t) list -> t -> string -> Yo
 val head_commit : t -> Repo_id.t -> Commit.t Current.t
 val refs : t -> Repo_id.t -> Commit.t Ref_map.t Current.Primitive.t
 val head_of : t -> Repo_id.t -> [ `Ref of string | `PR of int ] -> Commit.t Current.t
-val ci_refs : t -> Repo_id.t -> Commit.t list Current.t
+val ci_refs : ?staleness:(int option) -> t -> Repo_id.t -> Commit.t list Current.t
```
talex5, line 38: We already have a dependency on the [duration](https://github.com/hannesm/duration) library.
```suggestion
val ci_refs : ?staleness:(Duration.t option) -> t -> Repo_id.t -> Commit.t list Current.t
```

plugins/github/current_github.mli
```diff
@@ -96,8 +99,11 @@ module Api : sig
   (** [head_of t repo id] evaluates to the commit at the head of [id] in [repo].
       e.g. [head_of t repo (`Ref "refs/heads/master")] *)
 
-  val ci_refs : t -> Repo_id.t -> Commit.t list Current.t
-  (** [ci_refs t repo] evaluates to the list of branches and open PRs in [repo], excluding gh-pages. *)
+  val ci_refs : ?staleness:(int option) -> t -> Repo_id.t -> Commit.t list Current.t
```
talex5, line 102:
```suggestion
  val ci_refs : ?staleness:(Duration.t option) -> t -> Repo_id.t -> Commit.t list Current.t
```

talex5 (commented):
Thanks - this looks very useful! I've attached some minor comments / questions.

plugins/github/api.ml
patricoferris: Nope, will add, thanks :)

plugins/github/api.ml
patricoferris: It was left intentionally like that... It seems it's converting to the `Git` version of a commit id https://github.com/ocurrent/ocurrent/pull/231/files#diff-7928eb03e6563924a0c83bf56953954fa74e5689d4dcf80c02c03cdb4961b9b3R130 

So I thought it could just be left blank given it's about to be dropped by that function?

plugins/github/api.mli
patricoferris: Yep, it's okay that it is a transitive dependency right? Coming from `Current`?

avsm: Adding duration explicitly to the plugin package is fine too.

talex5:
Perhaps there should be a special case for the `master` branch. We should probably continue testing that even if it's stale.

patricoferris:
That sounds very reasonable -- do you think it think makes sense to query for the `defaultBranchRef` from the `repository` ([docs](https://docs.github.com/en/free-pro-team@latest/graphql/reference/objects#repository)) and then use that to check for the `master` equivalent (`trunk`, `main` etc.)?

talex5:
Yes, that sounds sensible!

patricoferris:
Tested the latest commit locally on a repo that has `trunk` as the default and seems to work (keeps it no matter the duration). Had to change the signature of `refs` function to return `ci_refs` which is current abstract for `Current_github` which makes it a little useless for end-users. Do you think I should make it non-absract or at the very least provide a function: 

```ocaml
val get_refs : ci_refs -> Commit.t Ref_map.t
```

To enable end-users to use it if they so wish?

talex5:
Yeah, there are several options here:
1. Make the type non-abstract.
2. Provide accessor functions.
3. Keep the old signature for the `refs` function, hiding the default branch for now, and worry about it later.

I'd be OK with any of the above.

talex5:
I've rebased this on master and made a few changes. In particular, I added staleness support to `Repo.ci_refs` too.

PR #233, opened 2020-12-04
Switch from "capnpc" to "capnp compile"

MisterDA:
`capnpc` seems to be the old name, and isn't present on Windows.


talex5:
Thanks!

PR #236, opened 2020-12-15
[POC] nix / opam2nix

timbertson:
First off, ocurrent seems really nice! I had fun putting this together.

For context, @avsm in a meeting the other week mentioned the possibility of integrating `opam2nix`-based builds in ocaml's CI system, to gain a good understanding of how many packages build without issues under nix. (aside: I'm now not sure whether he meant `opam-repo-ci` or `opam-ci`)

I don't know too much about it, but it was interesting enough to whip up this POC for building arbitrary nix derivations, and building opam2nix support on top of that. It currently only works on a single machine, i.e. it assumes `/nix/store` is shared between all jobs.

The resulting graph is interesting, although probably a little overwhelming. It starts simple:
Screenshot from 2020-12-13 21-55-06

Figures out how to build opam2nix:
Screenshot from 2020-12-13 22-08-35

And then gets rather carried away building `lwt` :grimacing: 
Screenshot from 2020-12-13 22-12-06

So yep. I've written a bunch of hacky code to get an idea of what's possible, and now it's time to find out more about how this could actually work in practice. I have many questions, I'm happy to discuss here or set up a meeting if that works better for you:

# About ocaml-ci:

### Will a single host run tasks concurrently?

I have some tasks which cannot be run concurrently on the same host since they'll end up fighting over a particular cache directory. Do I need to workaround that or is it fine?

### How longlived are the hosts?

I am doing some things that are efficient, but only if the user's ~/.cache directory is longlived. Should I expect that to be the case?

### How much should the pipeline be responsible for?

I'm currently assuming that it'd be OK to preinstall nix on the build hosts (or some bade image?) and just assume its presence, but that these should be dynamic and under the control of the pipeline:

 - nixpkgs (the core package definitions)
 - opam2nix
 - opam-repository

Actually installing nix during the pipeline is probably a bit of hassle, and likely won't have that much benefit (I don't imagine a need to test against multiple nix versions)

# Nix specifics:

### lifetime of /nix/store

I'm hoping the `/nix/store` could be made somewhat longlived. The more entries that already exist in there (per host), the less pressure there is on the online cache.

### Garbage collection

The jobs are currently not garbage collection-safe, i.e. doing a `nix-store --gc` might delete stuff which is currently in use, but not marked as a GC root.
I assume we wouldn't want `/nix/store/` to grow indefinitely, but perhaps garbage collection could run when no other (nix) jobs are running.

### Build graph stability:

The build graph, at least as I've implemented it, is quite unstable. That is:

 - a nix derivation logically depends on all of its build-time inputs
 - any nix derivation which is already present in a binary cache can short circuit this, i.e. I don't need the build dependencies of `foo` if I can just download the prebuilt `foo` from a cache

I've leveraged this to make a manageable dependency tree - a derivation already present in the binary cache has zero dependencies, and a derivation which actually needs building depends only on its direct dependencies which aren't available in a binary cache.
Without this shortcut, every derivation would "build the world", right down to bootstrapping GCC, coreutils etc.
The upshot is though, if you rebuild any of the "find dependencies" jobs you'll get an empty list because all the dependencies were built the first time round, and the dependency graph will kind of disappear.

I don't know if this is a problem in practice, but it definitely feels impure / odd.

### Collapsing sections

I would quite like the dependency graph of each derivation to be collapsible. But collapse seems to require a statically known key/value, which doesn't work for this (I'd want to use a key / value derived from the contents of a `string Current.t`). The graph seems to have node identities, is this something the framework could take care of?

### How to distribute secrets?

I haven't implemented it yet, but I'm planning to integrate cachix to handle multi-host builds. That is, building any derivation will also upload it to the binary cache, making it available for other builds. I don't know whether ocurrent has any better way of sharing files, but I figure a populated cache of prebuilt packages is probably a good side effect anyway :shrug:

I see how GITHUB_TOKEN works, is there any secret management that's available within a job though? Uploading build results to cachix would require some kind of key on the build host (not the process evaluating the pipeline).

# Random nits:

* I understood `gate` incorrectly - I thought `gate ~on:dep downstream` would prevent downstream from starting until `dep` had succeeded, but it seems to be more like `all`, in that it just waits for both to be complete. For me this was less useful, I ended up using a `bind` instead which seemed strange since I would prefer `downstream` to be displayed even before it was evaluated.


talex5:
Interesting!

The way things work with opam builds is that the OCurrent pipeline submits jobs to OCluster, which builds them on workers. The workers themselves do not run OCurrent.

We already have build types of `Dockerfile` and `OBuilder`. It would be quite easy to add `Nix build` as another type of build (see https://github.com/ocurrent/ocluster/blob/master/api/schema.capnp). Then we could have a Nix build pool with one or more machines running NixOS.

In that case, the OCurrent pipeline would run opam2nix as one step, and then submit it to the cluster in another step. I imagine the cluster could just do a normal nix build operation on it (the builds are sandboxed, right?).

It would be up to the build workers how they managed their caches, but there would need to be some kind of automatic pruning. I guess the Nix builds would be more efficient than the opam ones because they'd allow more reuse.

Note: `gate ~on:ctrl dep` does not delay `dep` itself, but can be used in place of `dep` to prevent `dep` from being passed to the next stage until `ctrl` is ready. e.g. if you want to prevent deploying a binary until the tests have passed:

```ocaml
let binary = build src in
let tests_ok = test binary in
binary |> gate ~on:tests_ok |> deploy
```

In other words, you're delaying the thing *after* the gate.

avsm:
This is fantastic! A few quick thoughts:
- we could generate a nix/obuilder build spec with a shared /nix/store. This might be the best of both worlds: we build in a controlled sandbox but use a shared cache. Might make GC easier as well as we could submit GC jobs after a particular run has finished.
- Can cachix be selfhosted? We have access to a big S3 bucket at the Scaleway site that runs the majority of our build VMs. Build secrets are just deployed with the infrastructure at the moment, so assuming it's in an environment variable is fine if we use the hosted version.

timbertson:
Thanks for the feedback, I'll start kicking the tyres of `ocluster` too :grinning: 

> In that case, the OCurrent pipeline would run opam2nix as one step, and then submit it to the cluster in another step. I imagine the cluster could just do a normal nix build operation on it (the builds are sandboxed, right?).

That's a good question. It might be more efficient if the scheduler could see into dependencies (to prevent multiple jobs all incidentally building a shared dependency like `lwt` in parallel), but that can probably be handled later.

> we could generate a nix/obuilder build spec with a shared /nix/store

Sounds good, but can you provide a little more detail? Does it conflict with @talex5 's suggestion of a different kind of `capnp` builder type for nix? Or would you use both of these approaches together (a new `nix` build type which executes within this obuilder image)

> Can cachix be selfhosted?

Not obviously, but I can ask about this when it comes time to the actual integration.

talex5:
> It might be more efficient if the scheduler could see into dependencies (to prevent multiple jobs all incidentally building a shared dependency like lwt in parallel)

Ah, I was assuming nix's builder would handle that. If not, we can certainly use our own builder (e.g. using your OCurrent code).

We could use a either a new build type or reuse OBuilder. Using OBuilder just means that the clients submitting jobs to the cluster are trusted not to mess up the nix cache, whereas a custom builder can enforce that itself. Also, we'd need a new kind of cache because at the moment OBuilder only allows one build to update a shared cache at a time (the others get a fork that is discarded after the build). There's no great difficulty adding that, though.

timbertson:
> Also, we'd need a new kind of cache because at the moment OBuilder only allows one build to update a shared cache at a time (the others get a fork that is discarded after the build). There's no great difficulty adding that, though.

Can you point to the cache code? The `cache_hint` makes it harder to find ;)
I see `state_dir`, but it sounds like that's more of a scratch directory than a shared cache.

What kind of shared cache is this? Is it an external S3 cache, or some kind of shared FS? If it's just s3, it's probably easier to use cachix since we'll want to populate build results there anyway.

> Using OBuilder just means that the clients submitting jobs to the cluster are trusted not to mess up the nix cache, whereas a custom builder can enforce that itself.

Can you lay out a little more explicitly what using OBuilder would mean? Would it effectively mean that a "nix build" action is "build this obuilder image", where the image is `<common nix obuilder preamble>; nix-store --realise /path/to/derivation`?

I'll need to embed a binary nix build instruction file (`.drv`) as part of the build instructions, can this be serialized in an obuilder spec or would I need to do some base64 and back? If I'm doing that, would it make more sense to just make it a first-class job type? I think it'll be important to validate that build instructions are valid and can't write to arbitrary store paths, so perhaps that seals the deal for needing an explicit build type anyway.

talex5:
> Can you point to the cache code? The cache_hint makes it harder to find ;)

`cache_hint` is part of OCluster, not OBuilder, so that probably isn't helping ;-) Try these:

https://github.com/ocurrent/obuilder#run
https://github.com/ocurrent/obuilder/blob/d257f1942fe058374068f4392bdafedc96fbbba1/lib/s.ml#L41

The cache is a local directory, currently using btrfs or zfs snapshots for atomic updates.

> Can you lay out a little more explicitly what using OBuilder would mean?

OBuilder is roughly equivalent to `docker build`. So: a base image and a list of shell commands to run in it. Each shell command is run in a `runc` sandbox and the filesystem is snapshotted at the end of each command.

Since we control it, we can add other features if they're useful, such as mounting a multi-writer shared cache directory into the container namespace.

But having "nix build" as another build type alongside "docker build" and "OBuilder build" seems very reasonable to me too!

timbertson:
OK, after a bit of a break I've got something working with a native Nix build type in ocluster:

https://github.com/ocurrent/ocluster/pull/112

It's working locally for me, and I believe all the bootstrapping avoids the need for nix to actually be installed on the host (it just needs basics like `tar`, `curl`, etc).

There's still some limitations, like the nix builder probably won't work correctly if you try to do concurrent nix builds in a single worker. But perhaps it's time for me to clean up the commits / diff for actual review? What sort of testing (both automated and manual, perhaps in your actual cluster?) do you think is appropriate before merging some of this?

talex5:
> I believe all the bootstrapping avoids the need for nix to actually be installed on the host

What does the bootstrapping do? Is this setting up Nix in a container, or for setting it up on a host that doesn't have it already? We have some Ansible scripts to configure the hosts and there's no problem installing Nix there.

According to https://nixos.org/manual/nix/stable/#sec-building-simple:

> Nix also performs locking. If you run multiple Nix builds simultaneously, and they try to build the same derivation, the first Nix instance that gets there will perform the build, while the others block (or perform other derivations if available) until the build finishes

As far as I can see, there's no problem just letting Nix handle all of this. It already does each build in a separate user account (as well as using container sandboxing when running on Linux).

It looks like your API is quite low-level (instantiating a Nix expression as one OCluster build, and then passing each Nix derivation as a separate job). I wonder if this will cause problems with queuing. e.g. a Nix build will first have to queue the instantiation, then (I assume) queue the derivations one by one, whereas non-Nix builds will submit the whole build job in one go. If each one involves queuing for a few minutes when busy, the Nix ones will end up taking much longer.

I also wonder about the shared cache. Normally the build machines are responsible for clearing their caches as necessary, but here the client has to know whether a result is still available before submitting the job. We might want to have multiple caches (e.g. all Scaleway machines share one cache, while packet.com/Equinix Metal machines share another).

How about having the client just submit the whole Nix expression (as with `nix-build`) and let the worker deal with everything else, rebuilding whatever is necessary for that machine?

talex5:
> What sort of testing (both automated and manual, perhaps in your actual cluster?) do you think is appropriate before merging some of this?

Merging it seems fairly low risk, once we have the final design. We can then run a test instance of ocaml-ci or opam-health-check to submit a load of Nix build jobs and check it's working before updating the main ocaml-ci to add Nix tests.

OBuilder has a self-test feature where it tries to build a simple spec file at start-up (and from time to time after that). If that fails, it doesn't join the cluster. That avoids broken or misconfigured machines producing lots of broken builds.

timbertson:
> What does the bootstrapping do? Is this setting up Nix in a container, or for setting it up on a host that doesn't have it already? We have some Ansible scripts to configure the hosts and there's no problem installing Nix there.

It's setting up a minimal nix inside a container, with this there's no need to ever install nix on the host.

> As far as I can see, there's no problem just letting Nix handle all of this. It already does each build in a separate user account (as well as using container sandboxing when running on Linux).

Installing nix for multi-user builds can potentially be difficult (I haven't done it recently myself, and it may require an interactive install). It's faster & easier (for me) to test that this setup process works reliably and portably, than to mess about with ansible, since its state is well-defined and it runs quickly.

And the build sandboxing is great, but it's not a sufficient security mechanism. e.g. it's trivial to exfiltrate a host file's contents if you know its path, even with sandboxing enabled. To be fully protected you'd also need restricted evaluation, which disables some things that opam2nix uses (like import-from-derivation and builtins.fetchgit).

IMO it's cleaner to just use a runc sandbox, then you don't have to know about all the ways you should lock down nix for building untrusted inputs.

> I wonder if this will cause problems with queuing. e.g. a Nix build will first have to queue the instantiation, then (I assume) queue the derivations one by one, whereas non-Nix builds will submit the whole build job in one go. If each one involves queuing for a few minutes when busy, the Nix ones will end up taking much longer.

Yep, that's definitely true. Is this likely to be a problem in practice? How saturated is the build cluster, and does it stay that way or are spikes pretty brief?

> I also wonder about the shared cache. Normally the build machines are responsible for clearing their caches as necessary, but here the client has to know whether a result is still available before submitting the job.

The other benefit of having nix fully within the sandbox is that you can run `nix-collect-garbage` (or just blow it away) at any time, since it's only serving the one builder. If we used multi-user builds we'd need to be much more careful about when we could run GC, and marking things as GC roots between building and uploading to the binary cache. Since "upload to cache" happens within the ocluster build, whenever a builder is idle there's nothing on-disk that it can't re-fetch from the cache.

> How about having the client just submit the whole Nix expression (as with nix-build) and let the worker deal with everything else, rebuilding whatever is necessary for that machine?

The main downside is wasted build time. In the case of a large-scale cache invalidation (like any commit to `opam2nix`), you'd end up with every single builder building ocaml, opam2nix, dune, etc from scratch. It does give nix builds a scheduling disadvantage to break things up, but if the cluster isn't saturated then it should end up doing far less work overall.

Oh and also: breaking it up by derivation means we attribute failures correctly. if `dune` fails to build, there's one build failure rather than 1000.

talex5:
> And the build sandboxing is great, but it's not a sufficient security mechanism. e.g. it's trivial to exfiltrate a host file's contents if you know its path, even with sandboxing enabled.

I hadn't realised that. How do you do it?

According to https://nixos.org/manual/nix/stable/:
> builds will be performed in a sandboxed environment, i.e., they’re isolated from the normal file system hierarchy and will only see their dependencies in the Nix store, the temporary build directory, private versions of /proc, /dev, /dev/shm and /dev/pts (on Linux), and the paths configured with the sandbox-paths option. This is useful to prevent undeclared dependencies on files in directories such as /usr/bin. In addition, on Linux, builds run in private PID, mount, network, IPC and UTS namespaces to isolate them from other processes in the system (except that fixed-output derivations do not run in private network namespace to ensure they can access the network).

I tried connecting into a build sandbox (using `nativeBuildInputs = [ breakpointHook ]`) and it seemed fairly isolated. e.g. `/etc/passwd` was a dummy file without my regular user listed. I see that the sandbox includes a full copy of `/nix/store`, not just the things I listed as dependencies, though, which seems not ideal. I suppose there shouldn't be anything very private there.

The reason I'm worried about this is that we have two protection systems here: the runc sandboxing to protect the host from Nix, and the Nix sandboxing to protect Nix from the builds (and the builds from each other). I guess we can only have one or the other enabled at a time, so we need to decide which is more important.

> To be fully protected you'd also need restricted evaluation, which disables some things that opam2nix uses (like import-from-derivation and builtins.fetchgit).

Yes, nix-instantiate needs to be sandboxed certainly. But that is separate from the store, right? If I understand correctly, nix-instantiate is normally run by the (untrusted) user, and submits build jobs to the trusted store, which builds them using the sandbox.

timbertson:
> The reason I'm worried about this is that we have two protection systems here: the runc sandboxing to protect the host from Nix, and the Nix sandboxing to protect Nix from the builds (and the builds from each other). 

Ah sorry, I should have mentioned that - I had to turn off nix's sandboxing because it didn't work under runc . Since we're already trusting runc for sandboxing obuilder, this seemed like the easiest approach to reason about, and gives a smaller attack surface overall.

> I hadn't realised that. How do you do it?

So yeah this is fun, you basically just need to cause the file to be read at evaluation time, not build time:

This won't work, because of sandboxing:

```nix
with import <nixpkgs> {};
stdenv.mkDerivation {
  name="secrets";
  buildCommand = "cat /etc/sysctl.conf > $out";
}
```

But this will:

```nix
with import <nixpkgs> {};
stdenv.mkDerivation {
  name="secrets";
  buildCommand = "cat ${/etc/sysctl.conf} > $out";
}
```

The distinction is that in the second case, you're using a nix path type, and ([hand wavey](https://shealevy.com/blog/2018/08/05/understanding-nixs-string-context/)) when that's used in a derivation, it copies that path into the store and the expression resolves to the now-imported nix store path This ensures the expression will be sensitive to changes in that file's contents. The _actual_ buildCommand after evaluation will be e.g. `"cat /nix/store/4wangrfsp695aybb2inqknwcgash54g3-sysctl.conf > $out"`, which the sandbox is fine with.

There are also (discouraged) builtins that can do similar cheeky stuff, which are disabled in restricted-eval mode. But yeah, it's a large surface area to have to think about.

> Yes, nix-instantiate needs to be sandboxed certainly

Sandboxing nix-instantiate would take care of the above, but running `nix-instantiate` within a sandbox and `nix-build` outside a sandbox (relying on build sandboxing) could have other inconsistencies, and (recurring theme) I'm not entirely sure it's enough, because of things like import-from-derivation which blur the line between instantiation and building.

Basically, there's just way less to think about if you just throw the whole thing in a sandbox ;)

talex5:
> Sandboxing nix-instantiate would take care of the above, but running nix-instantiate within a sandbox and nix-build outside a sandbox (relying on build sandboxing) could have other inconsistencies

But isn't this just how Nix works normally? When you install NixOS and run e.g. `nix-env` as a user this is exactly what happens - the `nix-env` runs as the unprivileged user and submits the derivations to the store using `nix-store`, which builds them using a Nix sandbox.

> and (recurring theme) I'm not entirely sure it's enough, because of things like import-from-derivation which blur the line between instantiation and building.

> Basically, there's just way less to think about if you just throw the whole thing in a sandbox ;)

But wouldn't that mean that users of the cluster can interfere with each others builds? OCluster is supposed to prevent that. In particular, someone submitting a PR to a repository would have their code run inside the `runc` sandbox, but with write access to the Nix store shared with other users, I think.

timbertson:
(apologies for the long delays between replies, I'm context switching on a few side projects :wink: )

> But isn't this just how Nix works normally? When you install NixOS and run e.g. nix-env as a user this is exactly what happens - the nix-env runs as the unprivileged user and submits the derivations to the store using nix-store, which builds them using a Nix sandbox.

Yes, that's true. And it's fine when you're running `nix-instantiate` as yourself. But doing this naively in ocluster would mean running `nix-instantiate` as the ocluster user, which is typicaly root. So you'd want _some_ sort of sandboxing beyond what happens for typical nix usage.

> But wouldn't that mean that users of the cluster can interfere with each others builds? OCluster is supposed to prevent that. In particular, someone submitting a PR to a repository would have their code run inside the runc sandbox, but with write access to the Nix store shared with other users, I think.

This is a good point. Reading [the nix manual](https://nixos.org/manual/nix/stable/#conf-build-users-group), I believe this can be fixed by using an unprivileged user in a shared user group and `1775` permissions on the /nix/store directory. That's getting close to reimplementing nix's own safety features at this point though.

I wonder, can any btrfs features you use for obuilder help? e.g. would it be easy to use some sort of copy-on-write to implement "discard all changes except those made to <a specific path>" after each build? Or the alternative "this whole directory is readonly, except you can write <this path that doesn't yet exist>"?

PR #245, opened 2020-12-31
Replace deprecated Fmt functions

MisterDA:
This PR is trivial as Fmt already aliases the deprecated functions to their replacement. They were only renamed.


talex5:
Thanks!

PR #246, opened 2021-01-04
Windows

MisterDA:
There are various places where I think that quoting the parameter with `%S` instead of `%s` is unnecessary.

### DIFF
--- a/lib/job.ml
+++ b/lib/job.ml
@@ -76,7 +76,7 @@ let log_path job_id =
 let id_of_path path =
   match Fpath.split_base path with
   | parent_dir, leaf ->
-    Fpath.(base parent_dir // leaf) |> Fpath.to_string |> Filename.chop_extension
+    Fpath.(base parent_dir // leaf |> segs) |> String.concat "/" |> Filename.chop_extension
 
 let run_cancel_hooks ~reason hooks =
   let rec aux () =
--- a/lib/log_matcher.ml
+++ b/lib/log_matcher.ml
@@ -124,7 +124,7 @@ let analyse_string ?job log_text =
     Some report
 
 let analyse_file ?job log_path =
-  let ch = open_in (Fpath.to_string log_path) in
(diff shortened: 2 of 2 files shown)


talex5:
I'm not keen on removing the quotes. For example `Invalid job ID ""` is much clearer than `Invalid job ID`.

MisterDA:
I've added the quotes back, I think we can safely merge the three patches.

lib/job.ml
```diff
@@ -61,7 +61,7 @@ let jobs_dir = lazy (Disk_store.state_dir "job")
 let log_path job_id =
   let open Astring in
   let jobs_dir = Lazy.force jobs_dir in
-  match String.cuts ~sep:"/" job_id with
+  match String.cuts ~sep:Fpath.dir_sep job_id with
```
talex5, line 64: This doesn't seem right. The syntax of job IDs shouldn't depend on the platform. Probably `id_of_path` should be changed so they always use `/`.

lib/job.ml
```diff
@@ -71,7 +71,7 @@ let log_path job_id =
       | Ok false -> Error (`Msg (Fmt.str "Job log %a does not exist" Fpath.pp path))
       | Error _ as e -> e
     end
-  | _ -> Error (`Msg (Fmt.str "Invalid job ID %S" job_id))
+  | _ -> Error (`Msg (Fmt.str "Invalid job ID \"%s\"" job_id))
```
talex5, line 74: Is this change intentional?

lib/job.ml
MisterDA: Changed to use `Fpath.segs` then re-concatenating the results with `/`.

lib/job.ml
MisterDA: I took to literally your "I'm not keen on removing the quotes". Removed the commit.

talex5:
Thanks!

Issue #247, opened 2021-01-27
Current_git.clone: default branch is not always master


TheLortex:
The optional parameter `gref` of `Current_git.clone` is set to `master` by default. As more and more repositories switch to other default branch names (such as `main`), I believe this optional parameter could have a better behavior when it's not set, by automatically figuring out the default branch of the repository instead of using `master`.

dra27:
This was discussed this morning. As of recent gits, assuming that your distro hasn't configured it directly (Git-for-Windows, for example, asks whether to use `master`, `main` or something else during its setup program and sets `init.defaultBranch` accordingly), then you get this message at `git init`:

```
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint:
hint:   git config --global init.defaultBranch <name>
hint:
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint:
hint:   git branch -m <name>
```
So although `main` is becoming quite common - and may even become the default - the point of the change from the Git tool's perspective is that they're _stopping_ hard-coding `master` everywhere and so mblessing "default is not `master`" as opposed to "default is `main`".

On that basis, perhaps deprecating the argument's optional nature is the better route to go - possibly with the call in #248 exposed as an API function so that a user of `Current_git` may choose to explicitly query the default branch (as opposed to implicit querying it by failing to specify an optional argument).

TheLortex:
> On that basis, perhaps deprecating the argument's optional nature is the better route to go - possibly with the call in #248 exposed as an API function so that a user of Current_git may choose to explicitly query the default branch 

Thanks, indeed that sounds like a good path

PR #248, opened 2021-01-27
Current_git.clone: find default branch when gref is not specified

TheLortex:
An attempt to fix https://github.com/ocurrent/ocurrent/issues/247

### DIFF
--- a/plugins/git/current_git.mli
+++ b/plugins/git/current_git.mli
@@ -40,7 +40,8 @@ module Commit : sig
 end
 
 val clone : schedule:Current_cache.Schedule.t -> ?gref:string -> string -> Commit.t Current.t
-(** [clone ~schedule ~gref uri] evaluates to the head commit of [uri]'s [gref] branch (default: "master"). *)
+(** [clone ~schedule ~gref uri] evaluates to the head commit of [uri]'s [gref] branch (default: make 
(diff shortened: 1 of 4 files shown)


plugins/git/current_git.mli
```diff
@@ -40,7 +40,8 @@ module Commit : sig
 end
 
 val clone : schedule:Current_cache.Schedule.t -> ?gref:string -> string -> Commit.t Current.t
-(** [clone ~schedule ~gref uri] evaluates to the head commit of [uri]'s [gref] branch (default: "master"). *)
+(** [clone ~schedule ~gref uri] evaluates to the head commit of [uri]'s [gref] branch (default: make 
+    use of `git ls-remote` to get the remote's default branch). *)
```
tmcgilchrist, line 44: At best the default git branch is mix of `master/main/trunk` the only reliable way to know which one to choose is to already know it or to query the Git hosting platform to tell you (GitHub/GitLab both have API calls to find this). Rather than making an educated guess here it would be better to keep the default as is, forcing an API change when you *must* supply a default branch seems unnecessary.

maiste:
Closing as now both `GitLab` and `GitHub` support the default branch request. Following @tmcgilchrist comment, it should be done with these APIs instead of `Git`.

Feel free to reopen.

PR #251, opened 2021-03-14
Update Dockerfile to latest opam-repository.

ewanmellor:
This is required because we depend on astring 0.8.5 (since 8df8b8530,
Dec 9 2020) and our previous pinned opam-repository doesn't include that
version of astring.  This implies that examples/docker_build_local has
been broken since then o_O.


talex5:
Thanks. Possibly we should get rid of this Dockerfile, and find something else to use for the example.

PR #253, opened 2021-04-06
Fixes

MisterDA:
Some little fixes (maybe more to come, who knows what the future is made of).


MisterDA:
hmm looking back at it, `conf-git` doesn't support Windows, neither does the dependency conf-which.

dra27:
This looks fine to me - the `conf-git` dependency isn't wrong, it just won't fix the issue on Azure (which is installing a _pre-compiled_ worker).

talex5:
Thanks!

PR #255, opened 2021-04-12
Implement labelling of clusters on the Graphviz diagram.

ewanmellor:
Add an optional label argument to list_iter, list_map, option_map.

Pass this through the List_map and Option_map metadata options to the
label of the subgraph in the Graphviz input.


talex5 (commented):
This looks nice. However, it doesn't compile on any version of OCaml... how are you testing this??

ewanmellor:
I have fixed and rebased.  I made the mistake of compiling as a submodule inside ocaml-ci, and I didn't realize that would miss a bunch of stuff inside ocurrent itself.  Apologies.

talex5 (commented):
Thanks. Testing it a bit more, it seems that the label is inherited by nested clusters. I assume that's not supposed to happen...

ewanmellor:
> Thanks. Testing it a bit more, it seems that the label is inherited by nested clusters. I assume that's not supposed to happen...

I updated this patch to explicitly set the label to the empty string on each unlabeled cluster. This stops the label from being inherited by subclusters.

talex5:
Thanks!

PR #256, opened 2021-04-12
Add a marker to each node in the diagram where there is a real job.

ewanmellor:
This makes it easy to distinguish those boxes where the user can click
through to get a build log.

### DIFF
--- a/lib_term/analysis.ml
+++ b/lib_term/analysis.ml
@@ -167,8 +167,22 @@ module Make (Meta : sig type t end) = struct
           | Error (_, `Msg msg) when error_from_self -> Some msg
           | _ -> None
         in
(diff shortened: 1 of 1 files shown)


talex5:
This doesn't compile (see errors from CI).

Styling clickable boxes sounds reasonable, but I'm not sure about these symbols. Pass/fail/active markers would be equally useful for boxes without logs, so I don't see why they would suggest that a box is clickable. Just adding `...` to all clickable boxes might work.

(Ideally, more of the boxes should be clickable. It would be useful to see logs for `monitor` boxes too, or be able to force a recheck of a job even if it doesn't produce logs.)

tmcgilchrist:
Merged as part of https://github.com/ocurrent/ocurrent/pull/333

PR #258, opened 2021-04-12
Add a new look-and-feel for the web UI.

ewanmellor:
This adds Roboto / Roboto Mono as web fonts, and adds highlight colors
for table heading and the nav menu based on the OCaml logo.  It adds the
logo itself to the nav bar and the favicon.
    
This adds a static module, which serves static files from a specified
docroot.  Hook this in to the routes under /images/*.
    
Lots of padding and spacing adjustments too.

Set the diagram node color (i.e. the box outline color) to match the fill.
This makes the boxes less heavy on the page, because they no longer have
the black border.
    

lib_web/static.ml
```diff
@@ -0,0 +1,21 @@
+let serve_file ~docroot uri =
+  let fname = Utils.Server.resolve_local_file ~docroot ~uri in
```
talex5, line 2: I think it would be better to list the resources explicitly here. In theory, `resolve_local_file` will deal with `..` and similar, but I'm not sure relying on that is a good idea. Especially on Windows, where someone might ask to read `CON` or something.

lib_web/context.ml
```diff
@@ -2,6 +2,14 @@ open Lwt.Infix
 
 let cookie_key = "__session"
 
+let img_ocaml_logo = "/images/ocaml%20logo.png"
```
talex5, line 5: Why should OCurrent pipelines have an OCaml logo? Performing CI for OCaml projects isn't the only use for pipelines.

lib_web/context.ml
```diff
@@ -72,15 +80,19 @@ let template t contents =
...
-          meta ~a:[a_charset "UTF-8"] ();
-        ]
+          link ~rel:[ `Icon ] ~href:img_ocaml_logo ();
+          meta ~a:[ a_charset "UTF-8" ] ();
+        ])
       )
       (body [
           nav [
+            img ~alt:"" ~src:img_ocaml_logo ~a:[ a_height 29 ] ();
+            div [code [txt site.name]];
```
talex5, line 94: The example apps do use the shell command as their name, but a pipeline's name is not "code" in general. Maybe just put a class on this and let the CSS decide what to do?

lib_web/context.ml
```diff
@@ -72,15 +80,19 @@ let template t contents =
...
             ul (
-              li [a ~a:[a_href "/"] [txt site.name]] ::
```
talex5, line 83: Normally, sites make the top-left logo/site-name go to the home page. If having two links is a problem, I suggest getting rid of the "Home" link instead.

talex5 (commented):
Thanks. The styling does need some work. The new design doesn't seem to work well with narrow windows, though.

Before (from ocaml-ci, running a slightly older version):

Screenshot_2021-04-13 ocaml-ci

After:

Screenshot_2021-04-13 docker_build_local

lib_term/dot.ml
```diff
@@ -30,7 +30,9 @@ let node f ?style ?shape ?bg ?url ?tooltip i label =
   let tooltip = Option.map (limit_str 4096) tooltip in (* (Graphviz max length is 16384) *)
   let attrs = [
     "label", Some label;
+    "color", bg;
     "fillcolor", bg;
+    "fontname", Some "Roboto, sans-serif";
```
talex5, line 35: It might be better to set the font globally, rather than duplicate it on every node. e.g. https://stackoverflow.com/a/10880578/50926

lib_web/current_web.mli
```diff
@@ -100,7 +100,7 @@ module Resource : sig
   end
 end
 
-val routes : Current.Engine.t -> Resource.t Routes.route list
+val routes : ?docroot:string -> Current.Engine.t -> Resource.t Routes.route list
```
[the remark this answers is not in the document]
ewanmellor: Added.

lib_web/style.ml
```diff
@@ -47,19 +106,27 @@ let css = {|
...
   }
 
   nav a {
+    display: block;
     color: #ddd;
     text-decoration: none;
-    padding-left: 0.5em;
-    padding-right: 0.5em;
+    padding: 2ex 0.75em;
+    transition: background 300ms, color 300ms;
```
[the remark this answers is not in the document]
ewanmellor: I've turned them into a kind of underline.  See if you like that better.

lib_web/context.ml
ewanmellor: I have renamed this to dashboard-logo.png.  Pipelines have full control over this image (because they control the docroot that images are served from) so with a generic name I think that all use cases are covered.

lib_web/context.ml
ewanmellor: I've changed this to a CSS class.

lib_web/context.ml
ewanmellor: I've put this link back.

lib_term/dot.ml
ewanmellor: I have done this.  There is now a new changeset in this PR, which adds a helper for writing the digraph stanza.  The fontname is then added at that point.

lib_web/static.ml
ewanmellor: `resolve_local_file` does indeed handle `..` and similar (I wrote the unit tests for that recently).

The next line in that code is passing the resolved filename to `Utils.Server.respond_file`, which is `Cohttp_lwt_unix.Server.respond_file`.  That function checks that the given filename resolves to a real file (`Unix.st_kind = S_REG`) so it will not serve special files.

Also, this is not serving arbitrary files even from the specified docroot, because it is under the control of `Current_web.routes`, which restricts the URL to `/images/*`.  I think that is sufficiently constrained.

ewanmellor:
> Thanks. The styling does need some work. The new design doesn't seem to work well with narrow windows, though.
> 
> Before (from ocaml-ci, running a slightly older version):
> 
> Screenshot_2021-04-13 ocaml-ci
> 
> After:
> 
> Screenshot_2021-04-13 docker_build_local
> 
> Also, CI is failing.

I have rebased, fixed the failing tests, and addressed all the other comments inline.

I have addressed the narrow-screen issue above by making the `query-form` class contents wrap at medium sizes, and then go to a single column form on mobile screens.

This is now ready for review again.

lib_web/static.ml
```diff
@@ -0,0 +1,21 @@
...
+  Utils.Server.respond_file ~fname ()
+
+(*
+  Serve files directly out of the given docroot.
+
+  Note that we ignore the given path, and we use the Context.uri instead.
+  The routes package has added a wildcard option
+  (https://github.com/anuragsoni/routes/pull/118) but this has not been
```
tmcgilchrist, line 10: This has been released as 1.0.0

lib_term/s.ml
```diff
@@ -109,21 +109,21 @@ module type TERM = sig
...
-  val list_iter : (module ORDERED with type t = 'a) -> ?collapse_key:string -> ('a t -> unit t) -> 'a list t -> unit t
+  val list_iter : (module ORDERED with type t = 'a) -> ?collapse_key:string -> ?label:string -> ('a t -> unit t) -> 'a list t -> unit t
   (** Like [list_map] but for the simpler case when the result is unit. *)
 
   val list_seq : 'a t list -> 'a list t
   (** [list_seq x] evaluates to a list containing the results of evaluating
       each element in [x], once all elements of [x] have successfully completed. *)
 
-  val option_map : ('a t -> 'b t) -> 'a option t -> 'b option t
+  val option_map : ?label:string -> ('a t -> 'b t) -> 'a option t -> 'b option t
```
tmcgilchrist, line 126: Could you add docs for `?label`?

PR #261, opened 2021-04-15
Log a message when cancelling a job finishes

talex5:
This avoids the confusing situation where the job finishes as you cancel, and you see the success message from the build as the last thing in the log, even though the job actually failed.

See e.g. https://github.com/ocaml/opam-repository/pull/18500#issuecomment-820447249

/cc @kit-ty-kate

### DIFF
--- a/lib_cache/current_cache.ml
+++ b/lib_cache/current_cache.ml
@@ -245,7 +245,9 @@ module Generic(Op : S.GENERIC) = struct
                   in
                   let outcome =
                     match Current.Job.cancelled_state op.job, outcome with
-                    | Error _cancelled, _ -> Error (`Msg "Cancelled")
+                    | Error (`Msg msg), _ ->
+                      Job.log job "%s" msg;
+                      Error (`Msg "Cancelled")
                     | Ok (), Ok _ -> Job.log job "Job succeeded"; outcome
                     | Ok (), Error (`Msg m) ->
                       Job.log job "Job failed: %s" m;


kit-ty-kate:
> where the job finishes as you cancel

I'm not sure this is what happens. From what I've been seeing this kind of failures are quite reliably triggered on large revdeps and those jobs just sit still after the last "saved as <hash>" (or just before i don't remember) until they timeout. If we have a successful job in the first place, couldn't we do the opposite and simply ignoring the cancellation if the job is actually done.

talex5:
Yes, I suspect there is some other problem causing the cancellation and success events to overlap (which would normally be quite unlikely). But this change might at least give a clue about what happened by including the final timestamp.

> If we have a successful job in the first place, couldn't we do the opposite and simply ignoring the cancellation if the job is actually done?

It's a bit risky. A typical case would be:

1. We ask a job to cancel.
2. The builder sends a TERM signal to the job.
3. The build script returns success.

It's hard to know whether the TERM signal had an effect (but the build script failed to report it), or whether the signal really had no effect (or was delivered after the build finished anyway). And if things aren't getting delayed then it won't normally make any difference.

( When it happens after the last OBuilder step is committed, it's probably waiting for btrfs to delete the working directory, and so this would be safe in the case of OCluster builds. Maybe the next btrfs update will fix everything... )

kit-ty-kate (approved):
> It's hard to know whether the TERM signal had an effect (but the build script failed to report it), or whether the signal really had no effect (or was delivered after the build finished anyway). And if things aren't getting delayed then it won't normally make any difference.

oh i see, fair enough. This change seems helpful indeed. I'm going to vendor it in opam-repo-ci right now and see if it helps us locate the issue. Thanks a lot!

kit-ty-kate:
Here is an example with this patch on: https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/6badcff2a0395ac33a501801b5b87cd7f6514ee8/variant/compilers,4.12,menhir.20210419,revdeps,posix-types.0.1.0

kit-ty-kate:
Copying the end of the log file here just in case it gets rebuilt:
```
-> installed posix-types.0.1.0
Done.
# Run eval $(opam env) to update the current shell environment
---> saved as "01497ff53ec16bc065768ca0b23e178ef7fdcaef561a340845f27c728d436605"
2021-04-19 08:42.00: Cancelling: Timeout (60.0 minutes)
Job succeeded
2021-04-19 08:50.00: Timeout (60.0 minutes)
```
The whole run had a very large amount of timeouts overall

talex5:
The timestamps on the files are (note: 1 hour ahead of CI timestamps):
```
-rw-r--r-- 1 root root  199 Apr 19 08:42 env       (7:42 in CI log)
-rw-r--r-- 1 root root 7144 Apr 19 09:00 log       (8:00 in CI log)
```
So it looks like:
- That job started at 8:42.
- It did the `btrfs subvolume snapshot` to start the last step at 08:59.
- It closed the log at 9:00, finishing the job, and snapshotted the result.
- At 9:42 (42 minutes later!) we asked it to cancel.
- Some time after that, OCluster reported `Job succeeded`.
- At 9:50 OCluster finally reported success to the client.

It looks like printing `---> saved as` is almost the last thing OBuilder does before returning to OCluster, so I have no idea where the delay is coming from.

Anyway, that's something to debug separately...

PR #263, opened 2021-04-29
Explicitly handle and diagnose EPERM when killing child processes.

ewanmellor:
If the job uses sudo (e.g. a developer using "sudo docker") then the job
will not be cancellable.  In this case, we get EPERM from the kill call.
Diagnose this with a more explicit message.  Also, do not kill the
main process in this case, but simply let the thread exit.

### DIFF
--- a/lib/job.ml
+++ b/lib/job.ml
@@ -96,7 +96,12 @@ let cancel t reason =
     Lwt.async (fun () ->
         Lwt.catch
           (fun () -> run_cancel_hooks ~reason hooks)
-          (fun ex -> Fmt.failwith "Uncaught exception from cancel hook for %S: %a" (id t) Fmt.exn ex)
+          (fun ex -> match ex with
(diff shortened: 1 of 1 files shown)


talex5:
This seems very ad-hoc. We can't hope to diagnose every possible error here. I suggest wrapping the code that sends the `kill`. Although the only place I can see that sends a kill already does that:

https://github.com/ocurrent/ocurrent/blob/b9dac6243d38b3ca8487fc6997fad3abeb88179d/plugins/docker/current_docker.ml#L74-L78

ewanmellor:
I wasn’t trying to diagnose every error. I was diagnosing the one that actually happens.

tmcgilchrist:
Merged as part of https://github.com/ocurrent/ocurrent/pull/333 to bring multicore-ci back onto ocurrent#master branch.

Is this change un-necessary @talex5 ?

PR #265, opened 2021-04-30
Select all checkbox and rebuild selected btn duplicated on top

MisterDA:


talex5 (approved):
Looks good! I pushed a small change to tidy the CSRF handling a bit - if you're happy with that then it's ready to merge.

talex5:
Note: the CI failure on Alpine can be ignored: it's because they promoted `capnproto-dev` from `testing` to `community`, but only `testing` has published the change so far.

MisterDA:
It's nicer indeed. Thank you for the change, I'm in favour of merging.

PR #266, opened 2021-05-05
ansi: 24-bit colour parsing

copy:
I'm using `ocurrent_ansi` to render terminal content live in the browser. For that I needed some extra ansi escape code handling, in particular 24-bit colours and the reverse attribute, which are implemented in this PR.

### DIFF
--- a/lib_ansi/escape_parser.ml
+++ b/lib_ansi/escape_parser.ml
@@ -1,12 +1,12 @@
 module Stream = Char_stream
 
 type colour =
-  [ `Black | `Blue | `Cyan | `Green | `Magenta | `Red | `White | `Yellow ]
+  [ `Default | `Black | `Blue | `Cyan | `Green | `Magenta | `Red | `White | `Yellow | `Rgb of int ]
 
 type sgr =
-  [ `BgCol of [ `Default | colour ]
+  [ `BgCol of colour
   | `Bold
-  | `FgCol of [ `Default | colour ]
+  | `FgCol of colour
   | `Italic
   | `NoBold
   | `NoItalic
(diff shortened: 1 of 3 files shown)


lib_ansi/escape_parser.ml
```diff
@@ -76,6 +92,7 @@ let read_intermediates ~params start =
     | Some (x, s2) when is_final_byte x -> (
         let func = Stream.(start -- s2 |> string_of_span) in
         let params = Astring.String.cuts ~sep:";" params in
+        let params = List.concat_map (Astring.String.cuts ~sep:":") params in
```
avsm, line 95: This use of List.concat_map pulls the supported OCaml version up by a couple. I think we're ok with OCaml 4.10+ for ocurrent, but @talex5 will confirm.

avsm:
> I'm using ocurrent_ansi to render terminal content live in the browser

This sounds pretty neat! Is it a publicly available project? I'm generally curious how you're finding using ocurrent in the browser.

copy:
> This sounds pretty neat! Is it a publicly available project? I'm generally curious how you're finding using ocurrent in the browser.

Not at the moment, but it's planned soon-ish. I'm only using current_ansi though. The terminal contents are collected from a local terminal (via [kitty](https://sw.kovidgoyal.net/kitty/)'s socket api), and can optionally be broadcast through a server.

talex5 (commented):
Looks good to me, once the opam constraints are updated so the CI is passing.

For consistency, we should update all the projects in the repository at once to require the new OCaml version (and also because the CI isn't very good at testing subsets of packages yet).

Or maybe we should just split `current_ansi` out into its own repository. There's no reason to release it in sync with the rest of OCurrent. Any objection to me doing that @avsm?

talex5:
Moved to new repository at https://github.com/ocurrent/ansi/pull/1.

PR #268, opened 2021-05-12
Don't pin current_ansi to this repo in Dockerfile

MisterDA:


talex5:
Thanks! We really should replace this Dockerfile with a simple example that doesn't need to be updated so much!

PR #276, opened 2021-06-23
add read and write functions to current_fs plugin

patricoferris:
Opening a draft PR to gather some feedback on extending the functionality of the FS plugin for OCurrent, in particular to add `File.read` and `File.write` functions where the former monitors the changes similarly to how `Current_git` monitors local repositories. 

One question I had was why `Current_fs` isn't a standalone plugin? As is, it is an optional library under `Current`, I couldn't see any use of `Current_fs.save` so maybe it makes sense to spin this into it's own thing so it doesn't add more dependencies to the core `Current` library?

There's a little example too that syncs a file to another one, you can do `dune exec -- examples/filesync.exe README.md README2.md` and then it should update `README2.md` if you make changes to `README.md`.

### DIFF
--- a/current.opam
+++ b/current.opam
@@ -31,6 +31,7 @@ depends: [
   "re" {>= "1.9.0"}
   "lwt-dllist"
   "alcotest" {>= "1.2.0" & with-test}
+  "irmin-watcher"
   "alcotest-lwt" {>= "1.2.0" & with-test}
   "astring" {>= "0.8.5"}
   "fpath" {>= "0.7.3"}
--- a/examples/dune
+++ b/examples/dune
@@ -7,7 +7,8 @@
                github
                github_app
                rpc_server
-               rpc_client)
+               rpc_client
+               filesync)
  (package current_examples)
  (libraries
    capnp-rpc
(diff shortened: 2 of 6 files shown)


talex5:
Yeah, improving this could certainly be useful.

It wasn't in its own package because it didn't add any dependencies. If it's adding irmin-watcher then it does need to be moved out.

Note: if you want to watch directories too then be careful: I have an idea that irmin-watcher is very slow in this case (I think it loads all the files in the watched subtree to check for changes). I'm not sure what happens if you watch an individual file - what happens if the file is replaced by a different one? You might need to watch the parent directory in that case. I have an idea this is why monitoring local Git repositories is so slow.

patricoferris:
Thanks for the useful feedback! Yeah I didn't realise that was the case in irmin-watcher (e.g. https://github.com/mirage/irmin-watcher/blob/master/src/backend.inotify.ml#L54-L58) thanks for pointing that out, I'll see if I can understand why that's the case because I feel like we really shouldn't need to do that :))

shonfeder:
Just cleaning up stale PRs. Please send word if you want to try to land this instead (or think it needs attention by someone else).

PR #277, opened 2021-07-07
Automatically refresh  jobs pages at fixed interval (60s)

MisterDA:
I find myself hitting the reload button sometimes... I thought that automatically refreshing the home and jobs pages every 60 seconds would be a good compromise to keep tracking the changes and not disturb the user too much if the page reloads while being used.

### DIFF
--- a/lib_web/context.ml
+++ b/lib_web/context.ml
@@ -67,15 +67,20 @@ let render_nav_link (link_label, path) =
   let open Tyxml.Html in
   li [a ~a:[a_href path] [txt link_label]]
 
-let template t contents =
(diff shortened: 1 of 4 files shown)


MisterDA:
Changed the PR to refresh only the jobs pages.

tmcgilchrist:
Could you add on pages like https://deploy.ci.ocaml.org/?repo=tmcgilchrist/opam2web ?

MisterDA:
> Could you add on pages like https://deploy.ci.ocaml.org/?repo=tmcgilchrist/opam2web ?

Done.

PR #283, opened 2021-07-23
Support Github rebuild via webooks.

tmcgilchrist:
Passing job_id through to set_status, which then gets sent back via webhooks and can be looked up as a job to rebuild.  The core piece is `CheckRunStatus` now includes the job_id
```ocaml
  type t = {
      state: state;
      description: string option;
      url: Uri.t option;
      job_id: string option;
    }
```

### DIFF
--- a/doc/examples/github.ml
+++ b/doc/examples/github.ml
@@ -44,12 +44,13 @@ let pipeline ~github ~repo () =
   |> Github.Api.Commit.set_status head "ocurrent"
 
 let main config mode github repo =
+  let has_role = Current_web.Site.allow_all in
   let engine = Current.Engine.create ~config (pipeline ~github ~repo) in
   let routes =
-    Routes.(s "webhooks" / s "github" /? nil @--> Github.webhook) ::
+    Routes.(s "webhooks" / s "github" /? nil @--> Github.webhook ~engine ~has_role) ::
     Current_web.routes engine
   in
-  let site = Current_web.Site.(v ~has_role:allow_all) ~name:program_name routes in
+  let site = Current_web.Site.(v ~has_role) ~name:program_name routes in
   Lwt_main.run begin
     Lwt.choose [
       Current.Engine.thread engine;
--- a/doc/examples/github.ml
+++ b/doc/examples/github.ml
@@ -35,7 +35,7 @@ let pipeline ~github ~repo () =
   let head = Github.Api.head_commit github repo in
   let src = Git.fetch (Current.map Github.Api.Commit.id head) in
   let dockerfile =
-    let+ base = Docker.pull ~schedule:weekly "ocaml/opam:alpine-3.12-ocaml-4.08" in
+    let+ base = Docker.pull ~schedule:weekly "ocaml/opam:alpine-3.13-ocaml-4.08" in
     `Contents (dockerfile ~base)
   in
   Docker.build ~pull:false ~dockerfile (`Git src)
@@ -47,7 +47,7 @@ let main config mode github repo =
   let has_role = Current_web.Site.allow_all in
   let engine = Current.Engine.create ~config (pipeline ~github ~repo) in
   let routes =
-    Routes.(s "webhooks" / s "github" /? nil @--> Github.webhook ~engine ~has_role) ::
+    Routes.(s "webhooks" / s "github" /? nil @--> Github.webhook ~engine ~webhook_secret:(Github.Api.webhook_secret github) ~has_role) ::
     Current_web.routes engine
   in
   let site = Current_web.Site.(v ~has_role) ~name:program_name routes in
--- a/doc/examples/github_app.ml
+++ b/doc/examples/github_app.ml
@@ -32,9 +32,9 @@ let weekly = Current_cache.Schedule.v ~valid_for:(Duration.of_day 7) ()
 
 (* Map from Current.state to CheckRunStatus *)
 let github_check_run_status_of_state ?job_id = function
-  | Ok _              -> Github.Api.CheckRunStatus.v ~url ?job_id (`Completed `Success) ~summary:"Passed"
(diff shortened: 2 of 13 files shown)


talex5 (commented):
This looks good. However, if we're performing actions in response to webhooks (rather than just triggering another query to GitHub) then we really should be validating incoming webhooks:

https://docs.github.com/en/developers/webhooks-and-events/webhooks/securing-your-webhooks

We should also think a bit about access control. Who does GitHub allow to click on the rebuild button? Possibly it should be tied in with the existing `has_role` access control system. At least, it needs to be documented.

tmcgilchrist:
> We should also think a bit about access control. Who does GitHub allow to click on the rebuild button? Possibly it should be tied in with the existing `has_role` access control system. At least, it needs to be documented.

Happily we get the GH User (as `sender` on the Webhook) that requested the rebuild so we can filter with `has_role`. We don't seem to have an option to not show the `Re-run` button for different users.
But we can provide our own set of `actions` as buttons on the CheckRun.

talex5 (commented):
I think the secret is needed for all web-hooks, not just for apps.

doc/examples/github_app.ml
```diff
@@ -31,14 +31,21 @@ let dockerfile ~base =
...
+  let+ md = Current.Analysis.metadata x
+  and+ state = Current.state x in
+  match md with
+  | Some { Current.Metadata.job_id; _ } -> github_check_run_status_of_state ?job_id state 
+  | None -> github_check_run_status_of_state state
 
 let pipeline ~app () =
   let dockerfile =
-    let+ base = Docker.pull ~schedule:weekly "ocaml/opam:alpine-3.12-ocaml-4.08" in
+    let+ base = Docker.pull ~schedule:weekly "ocaml/opam:alpine-3.13-ocaml-4.08" in
```
tmcgilchrist, line 48: Bumping this because we aren't publishing `alpine-3.12` since 4 months ago and it was causing issues running these examples.

plugins/github/current_github.ml
```diff
@@ -17,7 +17,12 @@ module Metrics = struct
     Counter.v_label ~label_name:"event" ~help ~namespace ~subsystem "webhook_events_total"
 end
 
-let webhook = object
+let validate_webhook_payload webhook_secret body headers =
+  let request_signature = Option.value ~default:"<empty>" (Cohttp.Header.get headers "X-Hub-Signature-256") in
+  let signature = "sha256=" ^ Hex.show @@ Hex.of_cstruct @@ Mirage_crypto.Hash.SHA256.(hmac ~key:(Cstruct.of_string webhook_secret) (Cstruct.of_string body)) in
+  String.equal signature request_signature
```
SquidDev, line 23: Does this need to use `Eqaf.equal` or some other constant-time string comparison function instead?

talex5: Indeed, the GitHub docs state:

> Using a plain == operator is not advised. A method like secure_compare performs a "constant time" string comparison, which helps mitigate certain timing attacks against regular equality operators.

https://docs.github.com/en/developers/webhooks-and-events/webhooks/securing-your-webhooks

doc/examples/github.ml
```diff
@@ -69,9 +70,18 @@ let repo =
     ~docv:"REPO"
     []
 
+let webhook_secret =
```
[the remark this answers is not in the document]
tmcgilchrist: This was only for the demo, the webhook secret from a file is here.
https://github.com/ocurrent/ocurrent/pull/283/files/4e0305335ff9d4feafae708c8926a1880766bdd6#diff-cdd21afd7356cc62801e8a3d1e29ed9791707d3f6639c817603b72cd21817ae3R210-R216

plugins/github/current_github.ml
tmcgilchrist: Thanks I didn't know about `Eqaf.equal` and that whole package.

doc/examples/github.ml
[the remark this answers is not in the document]
talex5: That link is broken now. But we want to make sure people read the secret from a file for OAuth pipelines too.

plugins/github/current_github.ml
```diff
@@ -17,7 +17,15 @@ module Metrics = struct
     Counter.v_label ~label_name:"event" ~help ~namespace ~subsystem "webhook_events_total"
 end
 
-let webhook = object
+let validate_webhook_payload webhook_secret body headers =
+  match webhook_secret with
+  | None -> true (* No webhook secret supplied skip the validation. *)
```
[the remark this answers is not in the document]
tmcgilchrist: Renamed this.

doc/examples/github.ml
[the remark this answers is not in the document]
tmcgilchrist: Changed the examples to read this from a file and restructured the types in Current_github.App and Api to include the web hook_secret. The examples now work the same as the other pipelines like `ocaml-ci`

plugins/github/current_github.mli
```diff
@@ -47,19 +56,19 @@ module Api : sig
...
 
     type state = [`Queued | `InProgress | `Completed of conclusion]
 
     val action: label:string -> description:string -> identifier:string -> action
 
-    val v : ?text:string -> ?summary:string -> ?url:Uri.t -> ?actions:action list -> state -> t
-    (** Construct a CheckRunStatus.t 
-        A maximum of three actions are accepted by GitHub.
+    val v : ?text:string -> ?summary:string -> ?url:Uri.t -> ?actions:action list -> ?identifier:string -> state -> t
+    (** [v ?text ?summary ?url ?actions ?identifier] creates a CheckRunStatus with [?text] description, a link to
```
talex5, line 70:
```suggestion
    (** [v ?text ?summary ?url ?actions ?identifier state] creates a CheckRunStatus with [?text] description, a link to
```

talex5 (approved):
Looks good to me!

PR #288, opened 2021-08-10
Provide markdown details for CheckRun.

tmcgilchrist:
Fill in more of the CheckRun payload to support actions, output markdown and Skipped conclusion type.

### DIFF
--- a/plugins/github/api.ml
+++ b/plugins/github/api.ml
@@ -94,36 +94,81 @@ module Status = struct
 end
 
 module CheckRunStatus = struct
-  type conclusion = [`Failure of string | `Success]
+  type conclusion = [`Failure of string | `Success | `Skipped of string]
   type state = [`Queued | `InProgress | `Completed of conclusion]
+  type action = { label:string; description:string; identifier:string }
   type t = {
       state: state;
-      description: string option;
+      summary: string option;
+      text: string option;
+      actions: action list;
       url: Uri.t option
     }
 
-  let v ?description ?url state =
-    let description = Option.map (Astring.String.with_range ~len:140) description in (* Max GitHub allows *)
-    { state; description; url }
+  let action ~label ~description ~identifier = {label; description; identifier}
 
-  let state_to_string = function
-    | `Queued -> 
-       ["status", `String "queued"]
-    | `InProgress -> 
-       ["status", `String "inprogress"]
-    | `Completed `Success -> 
-       ["status", `String "completed"
-       ;"conclusion", `String "success"]
-    | `Completed (`Failure msg) -> 
-       ["status", `String "completed"
-       ;"conclusion", `String "failure"
-       ;"output", `Assoc [("title", `String "Failure")
-                         ;("summary", `String msg)]]
+  let action_to_json {label; description; identifier} =
+    `Assoc [ ("label", `String label)
+           ; ("description", `String description)
+           ; ("identifier", `String identifier)]
 
-  let json_items { state; description; url } =
-    state_to_string state @
-    (match description with None -> [] | Some x -> ["description", `String x]) @
-    (match url with None -> [] | Some x -> ["details_url", `String (Uri.to_string x)])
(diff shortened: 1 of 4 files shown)


plugins/github/api.ml
```diff
@@ -94,35 +94,71 @@ module Status = struct
...
+  let action_to_json {label; description; identifier} =
+    `Assoc [ ("label", `String label)
+           ; ("description", `String description)
+           ; ("identifier", `String identifier)]
+
+  let v ?text ?summary ?url ?(actions=[]) state =
+    (* A maximum of three actions are accepted by GitHub. *)
+    { state; summary; text; actions; url }
+
+  let state_to_string description summary actions = function
```
talex5, line 119: This is badly named; it doesn't return a string!

talex5 (commented):
Looks OK to me.

The `state_to_string` function is a bit hard to check though. Perhaps it would be clearer written something like this:

```ocaml
  let state_json ~title ~summary ?description ~status () =
    let output = [ ("title", `String title)
                 ; ("summary", `String summary) ] @
                 (match description with None -> [] | Some x -> ["text", `String x])
    in
    [ "status", `String status
    ; "output", `Assoc output]

  let state_to_string description summary actions = function
    | `Queued -> 
      state_json ()
        ~title:"Queued"
        ~summary:(Option.value summary ~default:"Queued")
        ~status:"queued"
        ~description
    ...
```

plugins/github/api.ml
tmcgilchrist: Reworked this.

plugins/github/api.ml
```diff
@@ -94,36 +94,68 @@ module Status = struct
...
 
-  let json_items { state; description; url } =
-    state_to_string state @
-    (match description with None -> [] | Some x -> ["description", `String x]) @
-    (match url with None -> [] | Some x -> ["details_url", `String (Uri.to_string x)])
+    | `Completed `Success -> 
+      ("conclusion", `String "success") ::
+      state_json ~title:"Completed" 
+                 ~summary:(Option.value summary ~default:"Completed")
+                 ~status:"completed" ~text ()
```
talex5, line 144: This is OK, but note that doing it this way prevents using automatic indentation (e.g. with ocp-indent). The original style I suggested avoids that problem:
```ocaml
      state_json ()
        ~title:"Queued"
        ~summary:(Option.value summary ~default:"Queued")
        ~status:"queued"
        ~text
```

plugins/github/api.ml
```diff
@@ -94,36 +94,68 @@ module Status = struct
...
+
+    | `Completed (`Failure _) ->
+       [ "conclusion", `String "failure"
```
talex5, line 147: This could do with being run through an auto-indenter ;-)

plugins/github/api.ml
```diff
@@ -94,36 +94,68 @@ module Status = struct
...
+       ; "actions", `List (List.map action_to_json actions) ] @ 
+       state_json ~title:"Failure" 
+                  ~summary:(Option.value summary ~default:"Failure")
+                  ~status:"completed" ~text ()
+
+    | `Completed (`Skipped _) -> 
+      [ "conclusion", `String "skipped"
```
talex5, line 154: BTW, the idea with the `state_json` was to keep all the magic strings in one place. I just didn't read far enough to see that there were more. Perhaps it should have been:

```ocaml
  let state_json ~title ~summary ~text ~status ?conclusion ?(actions=[]) () =
    let output = [ ("title", `String title)
                 ; ("summary", `String summary) ] @
                 (match text with None -> [] | Some x -> ["text", `String x])
    in
    [ "status", `String status
    ; "output", `Assoc output ] @
    (if actions = [] then [] else ["actions", `List (List.map action_to_json actions)]) @
    (match conclusion with None -> [] | Some x -> ["conclusion", `String x])
```

talex5 (approved):
This looks fine. I made some more style comments, but these are just suggestions. Feel free to merge with or without them.

PR #291, opened 2021-08-20
Hide rules CRUD fields for non-logged in users.

tmcgilchrist:


talex5 (commented):
This is good, but it needs a couple more changes:
- Instead of explaining how to write a pattern, it should tell you to log in.
- If you don't need permission to add a rule, it should still show the boxes. e.g. when running the examples (which don't require authentication).

PR #297, opened 2021-10-04
Lwt_unix.yield was deprecated in favor of Lwt.pause

MisterDA:
Deprecated in the lwt dev, but it's a safe change to make now.

### DIFF
--- a/lib_web/log_rules.ml
+++ b/lib_web/log_rules.ml
@@ -36,7 +36,7 @@ let test_pattern pattern =
         begin
           if !i = 0 then (
             i := 100;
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
when was this function added. I suspect a lower bound bump might be needed

dra27:
It's at least in Lwt 3.0.0 and ocurrent already requires >= 4.3.0

MisterDA:
https://github.com/ocsigen/lwt/commit/f3c5fdb32c27be758c6ae6f42fd0762aaae47a06 before 2.1.0.

PR #302, opened 2021-11-11
Log extra context for Webhook validation failure.

tmcgilchrist:
Generates a prometheus metrics for failures that can be hooked into a
monitoring solution.

### DIFF
--- a/plugins/github/current_github.ml
+++ b/plugins/github/current_github.ml
@@ -17,11 +17,19 @@ module Metrics = struct
     Counter.v_label ~label_name:"event" ~help ~namespace ~subsystem "webhook_events_total"
 end
 
-let validate_webhook_payload webhook_secret body headers =
+let validate_webhook_payload webhook_secret body headers event =
   let request_signature = Option.value ~default:"<empty>" (Cohttp.Header.get headers "X-Hub-Signature-256") in
   let signature = "sha256=" ^ Hex.show @@ Hex.of_cstruct @@
     Mirage_crypto.Hash.SHA256.(hmac ~key:(Cstruct.of_string webhook_secret) (Cstruct.of_string body)) in
-  Eqaf.equal signature request_signature
+  if Eqaf.equal signature request_signature then
+    Ok ()
+  else
+    let s = Printf.sprintf {|
+Invalid X-Hub-Signature-256 received for %s, expecting %s == %s.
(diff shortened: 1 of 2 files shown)


talex5 (commented):
Looks useful - good idea putting a hint in the error message!

Note that we already have a Prometheus counter for "number of warnings logged". We should probably instead focus on making that one more useful, by removing bogus warnings. The main problem is we report pending installation requests as warnings. They should probably be reported separately as a guage.

plugins/github/current_github.ml
```diff
@@ -17,11 +17,19 @@ module Metrics = struct
...
 end
 
-let validate_webhook_payload webhook_secret body headers =
+let validate_webhook_payload webhook_secret body headers event =
   let request_signature = Option.value ~default:"<empty>" (Cohttp.Header.get headers "X-Hub-Signature-256") in
   let signature = "sha256=" ^ Hex.show @@ Hex.of_cstruct @@
     Mirage_crypto.Hash.SHA256.(hmac ~key:(Cstruct.of_string webhook_secret) (Cstruct.of_string body)) in
-  Eqaf.equal signature request_signature
+  if Eqaf.equal signature request_signature then
+    `Ok ()
```
talex5, line 25:
```suggestion
    Ok ()
```
Why not use the built-in type?

plugins/github/current_github.ml
```diff
@@ -30,15 +38,16 @@ let webhook ~engine ~webhook_secret ~has_role = object
     Log.info (fun f -> f "input_webhook: %a" Cohttp_lwt.Request.pp_hum req);
     let headers = Cohttp.Request.headers req in
     let event = Cohttp.Header.get headers "X-GitHub-Event" in
-    Log.info (fun f -> f "Got GitHub event %a" Fmt.(option ~none:(any "NONE") (quote string)) event);
-    Prometheus.Counter.inc_one (Metrics.webhook_events_total (Option.value event ~default:"NONE"));
+    let event_str  = Option.value ~default:"NONE" event in
+    Log.info (fun f -> f "Got GitHub event %a" Fmt.(quote string) event_str);
```
talex5, line 42:
```suggestion
    Log.info (fun f -> f "Got GitHub event %S" event_str);
```
I think this does the same thing.

talex5 (approved):

PR #303, opened 2021-12-03
Use opam 2.1 to build examples.

tmcgilchrist:
Also fixed some paths and added extra doc.

### DIFF
--- a/doc/examples/build_matrix.ml
+++ b/doc/examples/build_matrix.ml
@@ -1,3 +1,11 @@
+(* Usage: build_matrix.exe DIR
+
+   Given a local Git repository DIR, build the OCaml project with OCaml 4.10 and 4.11 on Debian.
+
+   e.g. Build the checked out ocurrent repository.
+   $ dune exec -- ./doc/examples/build_matrix.exe .
+*)
+
 let program_name = "build_matrix"
 
 open Current.Syntax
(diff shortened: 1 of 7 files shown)


doc/examples/build_matrix.ml
```diff
@@ -14,9 +22,9 @@ let dockerfile ~base ~ocaml_version =
   workdir "/src" @@
   add ~src:["*.opam"] ~dst:"/src/" () @@
   env ["OPAMERRLOGLEN", "0"] @@
-  run "opam install . --show-actions --deps-only -t | awk '/- install/{print $3}' | xargs opam depext -iy" @@
+  run "opam-2.1 install . --show-actions --deps-only -t" @@
   copy ~src:["."] ~dst:"/src/" () @@
-  run "opam install -tv ."
+  run "opam-2.1 install -tv ."
```
kit-ty-kate, line 27:
```suggestion
  run "ln -f /usr/bin/opam-2.1 /usr/bin/opam" @@
  run "opam init --reinit -n" @@
  add ~src:["*.opam"] ~dst:"/src/" () @@
  env ["OPAMERRLOGLEN", "0"] @@
  run "opam install . --show-actions --deps-only -t" @@
```
opam will upgrade the config and local state the first time it is run, however this takes time and is not backward compatible so it’s better to separate the upgrade from the rest to have a predictible base and to avoid issues with package that call opam itself (which would fail)

tmcgilchrist: Added a sudo for the `ln -f ` line but otherwise the same as suggested. Thanks @kit-ty-kate

PR #304, opened 2021-12-09
Set confirmation level

tmcgilchrist:
Explicitly set confirmation levels to allow for manually triggered jobs.

Pulling this across from @TheLortex fork for [tarides/tezos-ci](https://github.com/tarides/tezos-ci)

### DIFF
--- a/plugins/docker/build.ml
+++ b/plugins/docker/build.ml
@@ -4,6 +4,7 @@ type t = {
   pull : bool;
   pool : unit Current.Pool.t option;
   timeout : Duration.t option;
+  level : Current.Level.t option;
 }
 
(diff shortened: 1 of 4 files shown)


talex5 (approved):
Seems fine. Would be good to state in the PR what this is useful for, though.

TheLortex:
The goal is to implement CI jobs that can be run manually. Using this patch, it's possible to achieve that by setting them to a higher confirmation level than the other jobs.

TheLortex (approved):

TheLortex:
I believe there should be a similar patch in current_ocluster as well

MisterDA:
Please consider merging this! :)

PR #305, opened 2022-01-10
GitLab auth

tmcgilchrist:
https://github.com/tmcgilchrist/ocaml-gitlab/pull/30

### DIFF
--- a/plugins/gitlab/auth.ml
+++ b/plugins/gitlab/auth.ml
@@ -1,0 +1,123 @@
+open Lwt.Infix
+
+module Server = Cohttp_lwt_unix.Server
+
+type t = {
+  client_id : string;
+  client_secret: string;
+  scopes : string list;
+  redirect_uri : string;
+} [@@deriving yojson]
+
+let v ?(scopes=["read_user"]) ~client_id ~client_secret ~redirect_uri () =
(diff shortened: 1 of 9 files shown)


plugins/gitlab/auth.ml
```diff
@@ -0,0 +1,124 @@
...
+      | _, None -> Server.respond_error ~status:`Bad_request ~body:"Missing state" ()
+      | Some code, Some state ->
+        if state <> Current_web.Context.csrf ctx then (
+          Server.respond_error ~status:`Bad_request ~body:"Bad CSRF token" ()
+        ) else (
+          get_access_token t code >>= function
+          | None ->
+            Server.respond_error ~status:`Internal_server_error ~body:"Failed to get token" ()
+          | Some token ->
+            Log.warn (fun f -> f "Getting the current User");
```
talex5, line 81: Doesn't really sound like a warning to me.

talex5 (approved):
I'm not familiar with how GitLab auth works, but this looks sensible to me.

plugins/gitlab/auth.ml
tmcgilchrist: Very true, I'll remove it

PR #306, opened 2022-01-18
Be able to clone Git bare repository

dinosaure:
It's a simple patch which is currently used by our SMTP stack to be able to synchronize some stuffs with local Git repository made by `git init --bare`. /cc @TheLortex

### DIFF
--- a/plugins/git/clone.ml
+++ b/plugins/git/clone.ml
@@ -43,7 +43,7 @@ let build No_context job { Key.repo; gref } =
   end >>!= fun () ->
   Cmd.git_rev_parse ~cancellable:true ~job ~repo:local_repo ("origin/" ^ gref) >>!= fun hash ->
   let id = { Commit_id.repo; gref; hash } in
-  Lwt.return @@ Ok { Commit.repo = local_repo; id }
+  Lwt.return @@ Ok { Commit.repo = local_repo; id; bare = false }
 
 let pp f key = Fmt.pf f "git clone %a" Key.pp key
 
--- a/plugins/git/commit.ml
+++ b/plugins/git/commit.ml
@@ -11,6 +11,7 @@ end
 
 type t = {
   repo : Fpath.t;
+  bare : bool;
   id : Commit_id.t;
 } [@@deriving yojson]
 
--- a/plugins/git/current_git.ml
+++ b/plugins/git/current_git.ml
@@ -33,7 +33,7 @@ module Fetch = struct
       if Cmd.dir_exists local_repo then Lwt.return (Ok ())
       else Cmd.git_clone ~cancellable:true ~job ~src:remote_repo local_repo
     end >>!= fun () ->
-    let commit = { Commit.repo = local_repo; id = key } in
+            let commit = { Commit.repo = local_repo; id = key; bare = false } in
     (* Fetch the commit (if missing). *)
     begin
       Commit.check_cached ~cancellable:false ~job commit >>= function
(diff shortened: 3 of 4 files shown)


talex5:
Note: the [Git comment](
https://github.com/ocurrent/ocurrent/commit/7ed02003b142ea5e632eb58eefdef5b3fe2b3b1f) for the previous code says:
> copy local Git directories instead of cloning so that the pack files and timestamps don't change and mess up Docker's caching.

dinosaure:
Hmmhmm, the problem is that we need to clone _bare_ local repository in our case, do you have an other solution for that?

talex5:
Do you need the `.git` directory in the clone? If not, you could probably just `git checkout --work-tree=/repo/dir /tmp/dir`. There should probably be a separate `with_export` or something to do that anyway, as it would be faster when `.git` isn't needed.

dinosaure:
> Do you need the .git directory in the clone? If not, you could probably just git checkout --work-tree=/repo/dir /tmp/dir. There should probably be a separate with_export or something to do that anyway, as it would be faster when .git isn't needed.

Yeah we mostly interested by sources but I'm not the author of this commit and I don't have clues to "clone" a Git bare repository 😕 .

TheLortex:
Indeed we don't need the `.git` directory. I couldn't find any documentation for the `--work-tree` argument in `git checkout`, instead I'm reading about [git worktree](https://git-scm.com/docs/git-worktree): is it what you are referring to ?

I don't see a solution where we can have the full `.git` directory and keep good docker caching properties. So I'm in favor of using `git worktree` to checkout the given commit. In that situation the `.git` is a file with a deterministic content: `gitdir: XXX` pointing to the location of the tracked git directory. The solution also makes sense for non-bare directories and indeed might be more efficient. The `.git` is kept so it shouldn't break things that were relying on the git tool. What do you think ?

About the docker caching:
> copy local Git directories instead of cloning so that the pack files and timestamps don't change and mess up Docker's caching.

I agree that for pack files it's better to check out only what's needed, but for timestamps I believe changes don't invalidate the layer cache (see https://github.com/moby/moby/pull/12031)

talex5:
Not `git worktree` (that's something else). `git --work-tree` should apply to all commands - it's documented in `man git`. e.g.

```
mkdir /tmp/a
git --work-tree=/tmp/a checkout HEAD /tmp/a
ls /tmp/a
```

dinosaure:
What do you think about this new version? I did not test it.

talex5:
I think there should be two functions: one to clone with .git and one without (or it could be an argument to say which you want). If you only have one, then it's inconsistent.

dinosaure:
> I think there should be two functions: one to clone with .git

Ok but who decide which one to use? You expect a new argument which permits the user to choose if he/she wants to clone _via_ `checkout` or `clone`?

talex5:
I think the writer of the pipeline should know and say whether their build step requires access to `.git` or not.

dinosaure:
I rebased this PR with _upstream_, it's possible to have a review of it?

samoht:
Could you rebase again? And @talex5 what do you think about the PR?

dinosaure:
The pull-request was rebased.

talex5 (commented):
If I understand the PR correctly, when you ask to checkout a commit from a bare repository with the "Worktree" protocol then it:

1. Checks out the HEAD to the destination (without creating a `.git` directory).
2. Tries to `git reset` the destination to the desired commit.

Has this been tested? I don't see how it would work. Perhaps we should add some unit-tests for this, especially as the submodule stuff is difficult to get right.

Since this PR seems to be getting rather complicated, I suggest getting rid of the worktree branch. The `if bare` test means that normal users aren't affected by this change, so the only concern is that people specifically cloning a bare repository into a Docker container might see unnecessary changes to the .git subdirectory.

We could add a `?with_dotgit:bool` argument that simply deletes the `.git` directory at the end to avoid that.

PR #309, opened 2022-02-09
Show line numbers and allow jumping to specific lines in job logs

punchagan:
This PR adds some simple JS and CSS to allow jumping to a specific line number in the logs, and also displays line numbers for all logs

A URL like `http://localhost:8080/job/2022-02-09/115753-docker-build-71c777#L11-L15` would highlight lines 11 to 15 and scroll the page to line 11. 
image

### DIFF
--- a/lib_web/current_web.ml
+++ b/lib_web/current_web.ml
@@ -41,6 +41,7 @@ let routes engine =
     empty @--> Main.r ~engine;
     s "index.html" /? nil @--> Main.r ~engine;
     s "css" / s "style.css" /? nil @--> Style.r;
+    s "js" / s "line-numbers.js" /? nil @--> Script.line_numbers;
     s "pipeline.svg" /? nil @--> Pipeline.r ~engine;
     s "query" /? nil @--> Query.r ~engine;
     s "log-rules" /? nil @--> Log_rules.r;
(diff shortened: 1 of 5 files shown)


kit-ty-kate:
Could there be a special tag name for the last line as well by any chance?

kit-ty-kate:
> Some thing like `#end` always takes you to the last line of the log?

yes, that would be perfect! Thanks!

punchagan:
@art-w pointed out that the log response would be streamed by the server and this JS code wouldn't behave correctly when the output has not yet been completely rendered. I'll take a look into how to do this correctly and update the PR.

punchagan:
I've updated the PR to use a [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) to update the line numbers being displayed in the logs, as the logs get streamed from the server.

punchagan:
> > Some thing like `#end` always takes you to the last line of the log?
> 
> yes, that would be perfect! Thanks!

I've also added a commit that always scrolls the page to the end when `#end` URL hash is present.

punchagan:
@talex5 @tmcgilchrist could you please review this PR?

punchagan:
Thanks @tmcgilchrist !

PR #314, opened 2022-03-02
Pull collapse_list from TheLortex/ocurrent

tmcgilchrist:


TheLortex:
I don't know how useful is `collapse_list` in a general situation. 
In the documentation CI, it is used as a way to make a list of nodes collapsible while allowing individual nodes to be used in the rest of the pipeline.

This is solely to improve the pipeline graph rendering but we could also choose to drop that code and not use the pipeline rendering, building an new UI for that purpose instead

jonludlam:
Is that what we use for the `do/compile` phase? If so, none of that is currently working as I don't think you can expand the compile node on the live pipeline - presumably rendering the png times out or something.

TheLortex:
> Is that what we use for the do/compile phase? 

It's used here: https://github.com/ocurrent/ocaml-docs-ci/blob/main/src/pipelines/docs.ml#L85
We also use it in the prep phase and it's working fine.

PR #315, opened 2022-03-04
UI restyle

MisterDA:
Serve js and css from static files.


MisterDA:
Supersedes and closes #258.

samoht:
Could you rebase this to master?

tmcgilchrist (commented):
Can you rebase and try a deploy of it to ocaml-ci.

MisterDA:
I think this is ready, please let me know what you think.
The PR builds on the UI restyle by Ewan Mellor, fixes a few content-type bugs, serves files "statically" (which means they need to be installed on the host filesystem).
From the base ui-restyle PR, it removes the hovering animations and doesn't use the Roboto webfont, but rather the preferred system font.

Do you think static files should be served with something like ocaml-crunch instead? I think it's better either way not to embed them in the source, we get syntax highlighting and linting.

MisterDA:
This PR supersedes and closes #339.

tmcgilchrist:
Note that this needs updating in https://github.com/ocurrent/ocaml-multicore-ci/pull/22 so it can switch to mainline ocurrent.

lib_web/resource.ml
```diff
@@ -94,3 +94,42 @@ let logout = object
...
+          ("Content-Type", content_type);
+          ("Cache-Control", Printf.sprintf "public, max-age=%d;" max_age);
+        ]
+    in
+    Utils.Server.respond_string ~status:`OK ~headers ~body ()
+  end
+
+(* Serve a static asset from a resource embedded with ocaml-crunch,
+   with a cache of one day (by default). *)
+let crunch ?content_type ?(max_age=86400) _path = object
```
tmcgilchrist, line 117:
```suggestion
let crunch ?content_type ?(max_age=86400) _ = object
```

tmcgilchrist (approved):

MisterDA:
Applied your suggestion, thanks.

PR #317, opened 2022-03-10
Set the state dir following XDG Base Directory Specification

### LINKED ISSUE #106: Respect XDG_DATA_DIRS
The disk store directory is currently at `$PWD/var/`. 

It seems better to respect [`XDG_DATA_DIRS`](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) as a location for this cache, to avoid polluting the top-level directories of projects using OCurrent and needing to ignore `var/` everywhere.

The Dune source code contains an [internal XDG library](https://github.com/ocaml/dune/tree/master/src/xdg) that could be used as a starting point.

MisterDA:
Add a `--current-state-dir` option to set the state directory used
internally by OCurrent. The realpath of the executable is used to
derive a hash for the pipeline, to be able to re-use the same
directories across runs of the same pipeline.

Closes #106.

### DIFF
--- a/current.opam
+++ b/current.opam
@@ -26,6 +26,7 @@ depends: [
   "cmdliner" {>= "1.1.0"}
   "sqlite3"
   "duration"
+  "directories" {>= "0.5"}
   "prometheus"
   "dune" {>= "2.0"}
   "re" {>= "1.9.0"}
--- a/lib/config.ml
+++ b/lib/config.ml
@@ -24,13 +24,18 @@ let slow_start_thread t duration =
        )
     )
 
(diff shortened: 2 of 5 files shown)


tmcgilchrist:
> Would there be a cost to bumping to 4.13?

For discussion, I'd want our downstream pipelines to be migrated to use 4.13 in their deployment before forcing 4.13 in ocurrent. Following a very recent OCaml is ok since we control all the uses of it. There shouldn't be any reason we can't migrate but you'll need to do the legwork of bumping docker files and checking deployments. @MisterDA

PR #326, opened 2022-04-07
Set the current-state-dir via the cli

MisterDA:
Dropping XDG doesn't mean we cannot override the default state directory.
Does switching to a `ref` introduce an annoying side effect (if the caller changes the root state directory)? Should the variable be "read-only" once the cli switch overrides the default?

### DIFF
--- a/lib/config.ml
+++ b/lib/config.ml
@@ -3,6 +3,7 @@ open Lwt.Infix
 type t = {
   mutable confirm : Level.t option;
   level_cond : unit Lwt_condition.t;
+  state_dir : Fpath.t option;
 }
 
 let set_confirm t level =
@@ -24,9 +25,9 @@ let slow_start_thread t duration =
        )
     )
 
-let v ?auto_release ?confirm () =
+let v ?auto_release ?confirm ?state_dir () =
   let level_cond = Lwt_condition.create () in
-  let t = { confirm; level_cond } in
+  let t = { confirm; level_cond; state_dir } in
   Option.iter (slow_start_thread t) auto_release;
   t
 
@@ -35,6 +36,18 @@ let default = v ()
 let active_config : t option Current_incr.var = Current_incr.var None
 
 let now = Current_incr.of_var active_config
+
+let state_dir name =
+  let name = Fpath.v name in
+  assert (Fpath.is_rel name);
+  match Current_incr.observe now with
+  | None -> failwith "state_dir called before the state directory has been configured"
+  | Some { state_dir = None; _ } -> failwith "state_dir hasn't been configured"
+  | Some { state_dir = Some state_dir; _ } ->
+    let path = Fpath.append state_dir name in
+    match Bos.OS.Dir.create path with
+    | Ok (_ : bool) -> path
+    | Error (`Msg m) -> failwith m
 
 let rec confirmed l t =
   match t.confirm with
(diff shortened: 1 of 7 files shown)


lib/disk_store.ml
```diff
@@ -1,9 +1,9 @@
-let state_dir_root = Fpath.v @@ Filename.concat (Sys.getcwd ()) "var"
+let state_dir_root = ref (Fpath.v @@ Filename.concat (Sys.getcwd ()) "var")
```
talex5, line 1: Perhaps this should start as `ref None`, to make sure no-one uses the default before it gets set to the correct value?

talex5 (commented):
This seems reasonable to me. I keep thinking that the state directory shouldn't be a global at all, but changing that would likely be a lot of work.

lib/disk_store.ml
MisterDA: I think it's ok to leave it as-is, `Disk_store` isn't exposed and it's pretty clear that callers in `Current` should use `state_dir`.

> to make sure no-one uses the default before it gets set to the correct value?

The tests use the default value.

talex5: `state_dir` uses this, and therefore has the same problem, though. Making it an option would let `state_dir` fail if used too early.

MisterDA:
Ok, I've switched to `ref None` by default and made the `state_dir` parameter non-optional for the `Config.v` function.

lib/config.ml
```diff
@@ -65,8 +68,17 @@ let auto_release =
     ~docv:"SEC"
     ["confirm-auto-release"]
 
+let state_dir =
+  Arg.value @@
+  Arg.(opt string) (Filename.concat (Sys.getcwd ()) "var") @@
+  Arg.info
+    ~doc:"Set the state directory for the current pipeline."
+    ~docv:"DIR"
+    ["current-state-dir"]
```
talex5, line 77:
```suggestion
    ~doc:"Set the state directory for the pipeline."
    ~docv:"DIR"
    ["state-dir"]
```
(or use `ocurrent` instead of `current`, but none of the other options are prefixed this way)

lib/disk_store.ml
```diff
@@ -1,9 +1,9 @@
-let state_dir_root = Fpath.v @@ Filename.concat (Sys.getcwd ()) "var"
+let state_dir_root = ref None
 
 let state_dir name =
   let name = Fpath.v name in
   assert (Fpath.is_rel name);
-  let path = Fpath.append state_dir_root name in
+  let path = Fpath.append (Option.get !state_dir_root) name in
```
talex5, line 6:
```suggestion
  match !state_dir_root with
  | None -> failwith "state_dir called before the state directory has been configured"
  | Some state_dir ->
    let path = Fpath.append state_dir name in
```
It's quite likely users will hit this at some point.

MisterDA:
Applied you changes, thanks!

talex5 (commented):
I think this still won't quite work. `Config.default` immediately calls `Config.v`, which sets the (wrong) value. Other libraries that use it at initialisation time will then read the wrong value.

It could instead be a field on `Config.t`, read via `Config.now`. However, that would require changes in some users of the library, as they then need to wait for the configuration to happen. I think most places that use it do wait until the pipeline has started, as I always intended to fix this one day!

The web UI might cause a little trouble. e.g. it uses `Job.log_path`, which requires the state directory to have been set already, and it's not certain this will have happened by the time it's called, although it is very likely.

Doing this cleanly likely requires reworking a fair bit of code. `Current.state_dir` should disappear, being replaced by either `Config.state_dir` or `Engine.state_dir`, so it can't be called too early.

For a quick fix, another possibility is to use an environment variable rather than a command-line argument, since that can be read immediately.

MisterDA:
> Doing this cleanly likely requires reworking a fair bit of code.

Indeed. Just for the tests, it seems we have circular dependencies: the engine can't start if the `Git.reset ()` isn't called, which requires the configuration to have happened, but that's not the case since the engine hasn't started yet.

maiste:
@MisterDA was there a solution for the problem you raised, or should this PR be closed until we figure out what to do?

patricoferris:
Any updates on this? Just found I needed this for an ecology pipeline I've been building. For now I can shift the entire pipeline on the system to somewhere I have space for the engine's database but it would be nice to allow this to be given as an environment variable or programmatically set it :))

PR #327, opened 2022-04-07
Import and export rules from the web UI

MisterDA:
Allows to import (by a CSV file upload) and export rules (to a CSV
file) from the web UI. For now, uploading doesn't work as I'm not sure
how to correctly retreive the file data, and more importantly I can't
figure out why file file upload always errors with "Bad CSRF" (help
would be appreciated).

Sub-question: do we want to "atomically" import all rules, i.e. should
we validate the whole file before adding rules, or install rules but
retract installed rules if there's an error down the line, or just gather
errors as I've done?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+### Unreleased
+
+Web UI:
+
+- Allow to import and export using CSV the log rules. (@MisterDA, #327)
+
 ### v0.6
 
 Core:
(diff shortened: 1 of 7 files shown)


tmcgilchrist:
Importing atomically seems the right option and clearly report which rules failed the validation. 
e.g. regex invalid or duplicate rule

MisterDA:
The last commit should do the trick but I don't like much. It'd be better if resources could export the required role to access them, but I keep fighting the typesystem over objects.

tmcgilchrist (approved):

PR #328, opened 2022-04-11
Update to gitlab-unix.0.1.3

MisterDA:


current_gitlab.opam
```diff
@@ -21,7 +21,7 @@ depends: [
   "yojson"
   "cohttp-lwt-unix" {>= "4.0.0"}
   "dune" {>= "2.0"}
-  "gitlab-unix" {= "0.1.2"}
+  "gitlab-unix" {= "0.1.3"}
```
kit-ty-kate, line 24: I think it would be better if the constraints aren’t strict by default
```suggestion
  "gitlab-unix" {>= "0.1.3"}
```

tmcgilchrist (approved):

PR #330, opened 2022-04-15
Add cutoff operation

art-w:
This is a follow up on the ["abort propagation on equality" PR](https://github.com/ocurrent/ocurrent/pull/318) -- I now think that it can be very useful to customize the equality operator in some specific situations, as otherwise the code becomes terrible in order to conform to physical equality.

- I didn't want to add an `?eq` argument to `map`... its usage should be rare in practice, so the signature shouldn't encourage it :)
- The name `cutoff` and its documentation is rather low-level and might only speak to myself... your suggestions are welcome!

### DIFF
--- a/current.opam
+++ b/current.opam
@@ -19,7 +19,7 @@ doc: "https://ocurrent.github.io/ocurrent/"
 bug-reports: "https://github.com/ocurrent/ocurrent/issues"
 depends: [
   "ocaml" {>= "4.08.0"}
-  "current_incr" {>= "0.5"}
+  "current_incr" {>= "0.6.1"}
   "fmt" {>= "0.8.9"}
   "bos"
   "ppx_deriving"
--- a/lib_term/current_term.ml
+++ b/lib_term/current_term.ml
@@ -106,7 +106,7 @@ module Make (Metadata : sig type t end) = struct
       Current_incr.read y.v @@ Current_incr.write ~eq:(Dyn.equal ?eq)
     end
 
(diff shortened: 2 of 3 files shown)


talex5 (commented):
> I didn't want to add an ?eq argument to map... its usage should be rare in practice, so the signature shouldn't encourage it :)

Not sure about that. I think having `?eq` anywhere an equality test would be useful makes sense. In fact, I see that the `incr_map` function `map` uses already accepts this argument, so we probably just need to expose it in `map` itself.

art-w:
Ok sure, I'll add the `?eq` parameter: it's what [React](https://erratique.ch/software/react/doc/React/S/index.html) does, so there is a precedent! I liked the `cutoff` function as a way of keeping the other operations "simple looking" -- I still only have one use-case where this custom equality is required, and pipelines have been working without it for a long time now. The resulting code looked alright as a greppable warning for carefulness:

```ocaml
map ~eq:Custom.equal fn t  ===  map fn t |> cutoff ~eq:Custom.equal

my_pipeline |> cutoff ~eq:Custom.equal |> save_to_db
```

But I agree that the `?eq` syntax will be more useful for recurring/common datatypes :)

kit-ty-kate:
What is the state of this PR? Is this ready to be merged?

art-w:
Yes I just rebased it and updated the version bound on `current_incr` ! (for context, I believe that `opam-repo-ci` could benefit from this https://github.com/ocurrent/opam-repo-ci/pull/204 )

MisterDA:
Thanks!

Issue #334, opened 2022-04-19
Log matcher does not take the pattern with the highest score


kit-ty-kate:
```
# Current.Log_matcher.add_rule {pattern = "[\\n]test[\\n]"; report = "1"; score = 5};;
- : unit = ()
# Current.Log_matcher.add_rule {pattern = "[\\n].+[\\n]"; report = "2"; score = 1};;
- : unit = ()
# Current.Log_matcher.analyse_string "\ntest\n";;
- : string option = Some "2"
```

kit-ty-kate:
I think I found the function responsible:
https://github.com/ocurrent/ocurrent/blob/bf325ff53b2d834824e7e2af590fc74ebcb7f553/lib/log_matcher.ml#L94
`Re.Seq.all` used here does not return all the possible matches for the same part of the string, it only returns the most generic regexp apparently. It only returns the sequance of disjoint matches from what I understand.

PR #335, opened 2022-04-19
Log_matcher: Always match the pattern with the highest score first

kit-ty-kate:
Fixes https://github.com/ocurrent/ocurrent/issues/334

### DIFF
--- a/lib/log_matcher.ml
+++ b/lib/log_matcher.ml
@@ -20,7 +20,7 @@ module Dao = struct
                      report   TEXT NOT NULL, \
                      score    INTEGER NOT NULL, \
                      PRIMARY KEY (re))" |> Db.or_fail ~cmd:"create log_regexp";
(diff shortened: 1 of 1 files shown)


lib/log_matcher.ml
```diff
@@ -20,7 +20,7 @@ module Dao = struct
                      report   TEXT NOT NULL, \
                      score    INTEGER NOT NULL, \
                      PRIMARY KEY (re))" |> Db.or_fail ~cmd:"create log_regexp";
-    let load = Sqlite3.prepare db "SELECT re, report, score FROM log_regexp ORDER BY re" in
+    let load = Sqlite3.prepare db "SELECT re, report, score FROM log_regexp ORDER BY score DESC" in
```
talex5, line 23:
```suggestion
    let load = Sqlite3.prepare db "SELECT re, report, score FROM log_regexp ORDER BY score DESC, re" in
```
Just to keep it deterministic if two patterns have the same score.

kit-ty-kate:
All green and ready to merge.

PR #336, opened 2022-04-22
PRs: add informations

tmcgilchrist:
Found this change on @TheLortex 's fork being used in mirage-ci.


TheLortex:
Yes so that rationale is that we wanted more information when watching PRs, so `Ref.t` has been updated to provide a `pr_info` record.

The problem is that `head_of` takes a `Ref.t` for input so I have a second commit (in this branch https://github.com/TheLortex/ocurrent/tree/pr-info) to introduce a `Ref.id` with the original Ref type.

TheLortex:
Sorry for reviewing so long after you made your update. This looks good to me but the PR has to be rebased on top of master.

PR #345, opened 2022-05-04
Current_git: ensure stale submodules are removed

### LINKED ISSUE #340: Stale submodules not removed
CI for Eio is failing. The analyse step finds an opam file in the `ocaml-uring` submodule, which no longer exists:
```
2022-04-29 10:50.10: New job: Analyse
2022-04-29 10:50.10: Waiting for resource in pool analyse
2022-04-29 10:57.01: Got resource from pool analyse
2022-04-29 10:57.01: Checking out commit 6f6b2eee. To reproduce:
                       git clone --recursive "https://github.com/ocaml-multicore/eio.git" && cd "eio" && git fetch origin "refs/pull/212/head" && git reset --hard 6f6b2eee
2022-04-29 10:57.01: Exec: "cp" "-a" "--" "/var/lib/ocurrent/var/git/eio.git-2b23594b7c63b62fff46d7f80bbb69077b881604076ad7869a499d8832a61ef7/.git" 
                           "/tmp/git-checkout15a32123"
2022-04-29 10:57.01: Exec: "git" "-C" "/tmp/git-checkout15a32123" "submodule" 
                           "update" "--recursive"
Submodule path 'ocaml-uring': checked out '2ea06dbb9169964d328a36696907750ed6e85398'
Submodule path 'ocaml-uring/vendor/liburing': checked out 'ccfe96eb3dca697e58f1dc42f82fb79f01a948b1'
2022-04-29 10:57.02: Exec: "git" "-C" "/tmp/git-checkout15a32123" "reset" 
                           "--hard" "-q" "6f6b2eee4f1ee42697825309d3ac4b2e65da4eef"
warning: unable to rmdir 'ocaml-uring': Directory not empty
2022-04-29 10:57.02: Exec: "git" "-C" "/tmp/git-checkout15a32123" "submodule" 
                           "sync"
2022-04-29 10:57.02: Exec: "git" "-C" "/tmp/git-checkout15a32123" "submodule" 
                           "update" "--recursive" "--init"
2022-04-29 10:57.03: Exec: "find" "." "-maxdepth" "3" "-name" "*.opam"
...
2022-04-29 10:57.13: Results:
                       {
                         "opam_files": [
                           "eio.opam", "eio_linux.opam", "eio_luv.opam",
                           "eio_main.opam", "ocaml-uring/uring.opam"
                         ],
```
I'm not sure how it got into this state. OCurrent probably needs some unit-tests with submodules.

In any case, doing the `submodule update` before the `reset` is likely a mistake. Copying the [code in OCluster](https://github.com/ocurrent/ocluster/blob/1982b5dc07266631613263b32a67d2f88b37d39a/worker/context.ml#L69) might help.

talex5:
- Do `git submodule deinit` first, otherwise git won't remove old files on reset.
- Add a test for this.

I also removed the extra submodule update and sync added recently. The sync only affects initialised submodules and there aren't any with this change. The extra update was to help the sync succeed.

Fixes #340.

### DIFF
--- a/current_git.opam
+++ b/current_git.opam
@@ -28,6 +28,7 @@ depends: [
   "result" {>= "1.5"}
   "cstruct" {>= "6.0.0"}
   "mirage-crypto" {>= "0.8.0"}
+  "mdx" {with-test}
 ]
 build: [
   ["dune" "build" "-p" name "-j" jobs]
--- a/plugins/git/cmd.ml
+++ b/plugins/git/cmd.ml
@@ -59,5 +59,13 @@ let cp_r ~cancellable ~job ~src ~dst =
 let git_submodule_sync ~cancellable ~job ~repo =
   git ~cancellable ~job ~cwd:repo ["submodule"; "sync"]
 
+let git_submodule_deinit ~cancellable ~job ~repo ~force ~all =
+  let flags = List.concat [
+      (if force then ["--force"] else []);
+      (if all then ["--all"] else []);
+    ]
+  in
+  git ~cancellable ~job ~cwd:repo ("submodule" :: "deinit" :: flags)
+
 let git_submodule_update ~cancellable ~job ~repo ~init =
   git ~cancellable ~job ~cwd:repo ("submodule" :: "update" :: "--recursive" :: (if init then ["--init"] else []))
--- a/plugins/git/cmd.ml
+++ b/plugins/git/cmd.ml
@@ -36,8 +36,13 @@ let git ~cancellable ~job ?cwd args =
 let git_clone ~cancellable ~job ~src dst =
   git ~cancellable ~job ["clone"; "--recursive"; "-q"; src; Fpath.to_string dst]
 
-let git_fetch ~cancellable ~job ~src ~dst gref =
-  git ~cancellable ~job ~cwd:dst ["fetch"; "-f"; src; gref]
+let git_fetch ?recurse_submodules ~cancellable ~job ~src ~dst gref =
+  let flags =
+    match recurse_submodules with
+    | None -> []
+    | Some x -> ["--recurse-submodules=" ^ string_of_bool x]
+  in
+  git ~cancellable ~job ~cwd:dst ("fetch" :: flags @ ["-f"; src; gref])
 
 let git_checkout_force ~job ~repo treeish =
   git ~cancellable:false ~job ~cwd:repo ["-c"; "advice.detachedHead=false"; "checkout"; "-f"; treeish]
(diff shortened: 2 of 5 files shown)


MisterDA:
What if we keep the old code and also do `git clean -ffdx`? I agree that `git deinit` seems cleaner (and easier to understand) but maybe it's not as efficient cache-wise?

talex5:
I don't think deinit affects the caching:
> after deinitializing the Git directory is kept around
(https://git-scm.com/docs/gitsubmodules)

dra27:
Thanks for grasping this one! Does it still fix the original problem which is where the submodule URL _changes_?

talex5:
> I think from my experiments, git submodule deinit followed by a re-init/update still caused the wrong URL in the submodule’s git repo

Ah, you're right! If you deinit a submodule and then reinit it, it reuses the address from before the deinit instead of reinitialising it. How annoying.

I've added a second commit with a test for moving a submodule's repository and made some fixes for it. We now also fetch submodules during the `fetch` step, not when checking out (which might help with caching too).

talex5:
I also reverted "Checkout to the correct branch in with_checkout" from #333, since it prevents building PRs and therefore makes testing this difficult.

dra27:
I'm not sure that this `Dockerfile` is exactly what's happening now, but it demonstrates what I _think_ is still the problem:

```Dockerfile
FROM ocaml/opam AS base
RUN git clone https://github.com/ocurrent/ocluster.git --recursive
WORKDIR ocluster

FROM base
RUN sed -i -e s/ocurrent/dra27/ .gitmodules
RUN git submodule sync
RUN grep 'url = .*/obuilder\.git' * .git* -rI

FROM base
RUN rm -rf *
RUN git reset --hard HEAD
RUN sed -i -e s/ocurrent/dra27/ .gitmodules
RUN git submodule sync
RUN git submodule deinit --all --force
RUN git submodule update --init --recursive
RUN grep 'url = .*/obuilder\.git' * .git* -rI
```

In the first run, with the checkout as normal, we see:

```
.git/config:    url = https://github.com/dra27/obuilder.git
.git/modules/obuilder/config:   url = https://github.com/dra27/obuilder.git
.gitmodules:    url = https://github.com/dra27/obuilder.git
```

the second run attempts to simulate the default state of an OCurrent .git cache - which should be just a `.git` directory (because that's what we'd copied). At which point we have:

```
.git/config:    url = https://github.com/dra27/obuilder.git
.git/modules/obuilder/config:   url = https://github.com/ocurrent/obuilder.git
.gitmodules:    url = https://github.com/dra27/obuilder.git
```
`git submodule update` before the reset fixes it (which was the temporary fix). Another - possibly better - fix is to copy the `.git` directory _and any other `.git` files in the tree (so in the example, if `obuilder/.git` is restored after the `rm -rf` then the three URLs are also correctly updated).

talex5:
The URLs shouldn't matter in the checkout (after copying `.git`) because we're not doing any fetching at that point now (enforced with `~fetch:false`):

```ocaml
Cmd.git_submodule_update ~init:true ~cancellable:true ~fetch:false ~job ~repo:tmpdir
```

PR #350, opened 2022-06-02
gitlab: use set_status name (= context) in the web interface 

MisterDA:


plugins/gitlab/api.ml
```diff
@@ -269,7 +269,7 @@ let head_commit t repo =
 
 module Commit = struct
   module Set_status = struct
-    let id ="gitlab-set-status"
```
tmcgilchrist, line 272: Rather than having this hardcoded, you should be able to override the value with something meaningful for the application. Eg. for ocaml-ci this makes sense but for mirage-ci or tezos-ci it should be different.

tmcgilchrist (changes requested):
The API needs to change so you can set this value from a pipeline, rather than it being hardcoded.

tmcgilchrist (approved):

PR #352, opened 2022-06-08
Make line numbers clickable

MisterDA:


MisterDA:
Just re-tested this, and it works as I'd expect. It would also be nice later to have support for selecting parts of the logs.

PR #358, opened 2022-06-20
Add pull argument to docker compose

patricoferris:
This PR adds an optional `pull` argument to the Docker compose command and retains the previous behaviour by setting it by default to `true`.

I found this necessary if the compose file didn't actually use published images but instead used local image hashes. I don't know if that is a bit of a docker-compose anti-pattern, but in a pipeline scenario it might be somewhat common when people don't want to publish images?

### DIFF
--- a/plugins/docker/compose.ml
+++ b/plugins/docker/compose.ml
@@ -1,6 +1,6 @@
 open Lwt.Infix
 
-type t = No_context
+type t = { pull : bool }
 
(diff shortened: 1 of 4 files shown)


tmcgilchrist (approved):

maiste (approved):
LGTM :+1: 
Is it ok to rebase before merging? (I can force push the change if needed).

PR #363, opened 2022-06-25
Add build context parameter to docker build

patricoferris:
Currently the `Current_docker.Default.build` function allows a user to specify the relative path to a Dockerfile from the Git src, but there's no way as far as I can tell to specify the build context path. So if I have a Dockerfile in `src` that assumes the build context is `src` I can't run a build in that context.

The current implementation does potentially allow a user to escape the build directory though. I don't know if this is too much of a problem if the user is always specifying this path and maybe it is up to the user to check the `~path` variable is not trying to do something bad?

### DIFF
--- a/plugins/docker/s.ml
+++ b/plugins/docker/s.ml
@@ -56,7 +56,8 @@ module type DOCKER = sig
       @param dockerfile If present, this is used as the contents of the Dockerfile.
       @param pull If [true], always check for updates and pull the latest version.
       @param pool Rate limit builds by requiring a resource from the pool.
-      @param path The relative file path passed to the docker build command as the build context. *)
+      @param path The relative file path passed to the docker build command as the build context.
+                  No checks are done over the path: it can point anywhere outside the build directory. *)
 
   val run :
     ?label:string ->
--- a/plugins/docker/s.ml
+++ b/plugins/docker/s.ml
@@ -44,6 +44,7 @@ module type DOCKER = sig
     ?squash:bool ->
     ?label:string ->
     ?dockerfile:[`File of Fpath.t | `Contents of string] Current.t ->
+    ?path:Fpath.t ->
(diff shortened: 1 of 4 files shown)


maiste (approved):
LGTM :+1: 
@patricoferris, is it ok if I force-push your PR (after a rebase on master) before merging?
(If you allow edit by maintainers)

patricoferris:
Thanks for the review :)) Fine by me !

maiste:
I rebased on master. However, I prefer to have a second point of view for the escaping `path` issue before merging something that doesn't fit with the `ocurrent` team's point of view. As you question, do we want the user to have the possibility to escape from the build directory and make the user responsible for that or not?

talex5:
> As you question, do we want the user to have the possibility to escape from the build directory and make the user responsible for that or not?

Who is the "user" here? The author of the pipeline, or the person submitting the PR?

The case we want to prevent is someone making a PR in which `src` is a symlink to `/root` or something.

maiste:
In this case, can we consider this is something that `ocurrent` doesn't have to take care of as it will be used in the pipeline code?
If yes, maybe it is worth adding a warning about it in the documentation.

maiste:
I rebase over master and add a small warning. Do you think we can merge it?

plugins/docker/s.ml
```diff
@@ -28,7 +28,8 @@ module type DOCKER = sig
                       has no [valid_for] limit then we will only ever pull once. *)
 
   val peek :
-    ?label:string ->
+    ?label:string ->Some external check might be required
+                  to ensure it doesn't escape the build directory.
```
tmcgilchrist, line 32: Should this be a comment?

maiste: It shouldn't even be here. Thanks.

PR #374, opened 2022-09-01
Add Current.observe to inspect the state of a term

TheLortex:
This PR adds a function to enable looking at an ocurrent node's state.

I have extended the `'a Current.Output.t` result type to provide an additional `` `Blocked`` error state, happening when the node has failed because an upstream dependency failed. 

There are some tests to demonstrate the 4 possible cases.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+### dev
+
+API:
+
+- Add `Current_term.observe` to evaluate the current state of a term. (@TheLortex, #374)
+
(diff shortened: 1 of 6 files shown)


TheLortex:
Context: we use that in the monitoring page of docs CI https://github.com/ocurrent/ocaml-docs-ci/pull/76

MisterDA (approved):

MisterDA:
Thanks!

Issue #376, opened 2022-09-22
Update `Current_github` to support specific handler


maiste:
In the current state of the `webhook` function, the user cannot influence the way `Current_github` has to deal with the events it receives from *GitHub*. We should provide a way to give an `event handler` to make it possible to decide how to handle them from a user perspective.

tmcgilchrist:
The web hook handler is tightly coupled with the Github plugin machinery. Users providing their own function could break the ocurrent machinery. What use case are you trying to support?

maiste:
The case I wanted to cover here was the one where you want to execute different actions when receiving a pull request for example.
That's why I wanted to introduce a new function and not update the old one. It lets the user having a default version.

maiste:
Considering we have found other ways to tackle the problem, I'm closing this issue.

PR #377, opened 2022-09-22
GitHub: webhook with user handler

maiste:
This PR addresses #376. The goal is to allow users to use their handler with webhook as the default one is tight to CI behaviours.

### DIFF
--- a/plugins/github/current_github.ml
+++ b/plugins/github/current_github.ml
@@ -33,7 +33,7 @@ See https://docs.github.com/en/developers/webhooks-and-events/webhooks/securing-
     Error s
 
 
-let webhook ~engine ~get_job_ids ~webhook_secret = object
+let webhook_with_handler ~engine ~get_job_ids ~webhook_secret ?handler () = object
   inherit Current_web.Resource.t
 
   method! post_raw _site req body =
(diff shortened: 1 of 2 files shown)


tmcgilchrist:
Rather than adding a new api function, expose the default handler as a function and the function to provide a handler which defaults to using this function. eg

``` ocaml
val default_handler : ....

val webhook_with_handler : ?(handler = default_handler) ...
```

Clearly this breaks the current API version and it should be called out in the [CHANGES.md](https://github.com/ocurrent/ocurrent/blob/master/CHANGES.md) file.

tmcgilchrist (changes requested):
Please consider the requested API changes.

maiste:
The reason I was considering adding a new function was to extend the API and don't break the old one with `webhook` as it is used in various projects...

 In this case shouldn't we the API compatible with older versions?

novemberkilo:
I'm not sure if this is what @tmcgilchrist meant but perhaps we can keep the existing API by just doing:

```
val default_handler : ....

val webhook : ... ?(handler = default_handler) ...
```

So we would only introduce an optional argument to the existing `webhook` function that will default to default_handler. If I am following along correctly then this will not add any new functions to the API, retain backward compatibility, and also provide the ability for a handler to be supplied by a user of `webhook`

maiste:
As this might not be required, should I keep the version with no breaking change or introduce the one with a breaking change but keep this PR opens as a draft until we figure out if it's needed?

MisterDA:
IMO it's mostly ok to break the API if you submit PRs to all user, which we _can_ do because we're the users. And I've been doing that quite often.

maiste:
Closing. See #376

PR #380, opened 2022-10-13
Add directory source for docker builds

tmcgilchrist:
Used in https://github.com/ocurrent/macos-infra/pull/22

### DIFF
--- a/plugins/docker/build.ml
+++ b/plugins/docker/build.ml
@@ -17,7 +17,7 @@ let use_pool pool f =
 
 module Key = struct
   type t = {
-    commit : [ `No_context | `Git of Current_git.Commit.t ];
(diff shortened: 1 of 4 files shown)


maiste:
Isn't this PR a duplication of https://github.com/ocurrent/ocurrent/pull/363?

tmcgilchrist:
The difference is for macos-infra usage we are working out of a non git directory that is local to the build machine. For macOS-infra we are creating a dummy docker image, which contains macOS executables for an OCaml compiler install, but is not a real executable docker container since there is no native macos docker.

While for #363 wants to specify the directory (within a git checkout) to run the Dockerfile in.

maiste (approved):
Oh I see! Thanks.
LGTM 👍

PR #382, opened 2022-10-19
Disable uneeded tooltips on diagrams nodes

MisterDA:
When the tooltip [1] attributes isn't specified in graphviz files, the browser will generate a tooltip from the label or the node name which is unhelpfull. There is no way of explicitely disabling the tooltip apart from setting is to a space, which the browser don't display.

[1]: https://graphviz.org/docs/attrs/tooltip/


tmcgilchrist:
@MisterDA is this how we end up with tooltips named `n241` and other such wonderful names?

MisterDA:
> @MisterDA is this how we end up with tooltips named `n241` and other such wonderful names?

Exactly! With this patch the browser won't show tooltips for these nodes.

PR #383, opened 2022-11-02
Added ref_title accessor for commits (ocaml-ci)

benmandrew:
Currently the ref title of a commit is not exposed, which is necessary for displaying the list of refs retrieved for a repository via the `ci_refs` function for display in the ocaml-ci web UI. This adds an accessor for that field.

Note: while the gref is already provided which allows us to derive the title for a branch ('refs/heads/[title]'), this doesn't work for PRs which are identified by the ID number ('refs/pull/[id]/head'), thus necessitating the change.

### DIFF
--- a/plugins/github/api.ml
+++ b/plugins/github/api.ml
@@ -338,7 +338,7 @@ module Commit_id = struct
       match Astring.String.cuts ~sep:"/" gref with
       | "refs" :: "heads" :: branch ->
         Some (Astring.String.concat ~sep:"/" branch)
-      | _ -> Some gref)
+      | _ -> None)
     | `PR _ -> None
 end
 
--- a/plugins/github/api.ml
+++ b/plugins/github/api.ml
@@ -327,14 +327,19 @@ module Commit_id = struct
 
   let digest t = Yojson.Safe.to_string (to_yojson t)
 
-  let ref_title t =
+  let pr_name t =
+    match t.id with
+    | `Ref _ -> None
+    | `PR { title; _ } -> Some title
+
+  let branch_name t =
     match t.id with
     | `Ref gref -> (
-        match Astring.String.cuts ~sep:"/" gref with
-        | "refs" :: "heads" :: branch ->
-          Astring.String.concat ~sep:"/" branch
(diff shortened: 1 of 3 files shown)


plugins/github/api.mli
```diff
@@ -31,6 +31,7 @@ module Commit : sig
   val compare : t -> t -> int
   val set_status : t Current.t -> string -> Status.t Current.t -> unit Current.t
   val uri : t -> Uri.t
+  val ref_title : t -> string
```
tmcgilchrist, line 34:
```suggestion
  val pr_title : t -> string option
```
What about this? The title only makes sense for a Pull Request.

maiste: We also want the name of the branch: we might need another name...

benmandrew:
I have split the function for returning the ref names into two, one for PRs and one for branches. They return option types so that when ocaml-ci has the complete list of `ci_refs` it can call `filter_map` twice to partition the list into PRs and branches.

plugins/github/api.ml
```diff
@@ -327,14 +327,19 @@ module Commit_id = struct
...
     | `Ref gref -> (
-        match Astring.String.cuts ~sep:"/" gref with
-        | "refs" :: "heads" :: branch ->
-          Astring.String.concat ~sep:"/" branch
-        | _ -> gref)
-    | `PR { title; _ } -> title
+      match Astring.String.cuts ~sep:"/" gref with
+      | "refs" :: "heads" :: branch ->
+        Some (Astring.String.concat ~sep:"/" branch)
+      | _ -> Some gref)
```
maiste, line 341: To make it consistent with the name:
```suggestion
      | _ -> None)
```

benmandrew: Good catch, I missed that

maiste (approved):
Good on my side, thanks!
WDYT @tmcgilchrist?

tmcgilchrist:
Looks good to me.

maiste:
That's good to merge as soon as the CI goes green. We should release once this is merged to include the change with the route.2.0.0 available.

PR #387, opened 2022-11-18
Added ci_refs_with_default (ocaml-ci)

benmandrew:
If we get the default ref by calling `refs t repo |> default_ref` then we will unnecessarily duplicate calls to the GitHub API, this PR allows them to be aggregated.


benmandrew:
I missed `Github.Api.Repo.head_commit`, which provides the required functionality.

PR #388, opened 2022-11-27
[RFC]: Initial port to eio

patricoferris:
This is a draft PR to start porting the library to eio.

The tests are passing locally for me but I haven't ported some of the larger pieces (because of the effort to port their dependencies). So currently only the core `Current` library along with the Slack, FS, Docker and Git plugins are ported just to get an idea for what it would look like (I haven't tested the plugins though!). There are a few questions that it raises in my mind whilst porting. 

1. It seems there's a need for a "lifetime of the pipeline"/"lifetime of the `Eio_main.run`" switch (the `Eio.Switch.t` flavour) -- I've made it so this gets passed around a lot but I think it might be easier to do this differently. Maybe even the pipeline thunk should actually be `Eio.Switch.t -> unit Current.t` or something like that, or we have some effect to fetch the "Engine" switch which would mean the API doesn't need changed as much. 
2. I'm pretty sure the `read` and `write` functions for monitors are not quite right, I think there's a cleaner way to keep cancellation in the library compared to how I've done it in a few places.
3. The exact change/propagate semantics confused me slightly converting `Lwt.pause` to `Fiber.yield ()`. It seems to work but 🤷 
4. There's no nice way it seems to support redirecting stderr to stdout in luv/libuv so I've used a socket pair... probably not the most efficient ^^"


talex5:
> It seems there's a need for a "lifetime of the pipeline"/"lifetime of the Eio_main.run" switch (the Eio.Switch.t flavour) -- I've made it so this gets passed around a lot but I think it might be easier to do this differently. Maybe even the pipeline thunk should actually be Eio.Switch.t -> unit Current.t or something like that, or we have some effect to fetch the "Engine" switch which would mean the API doesn't need changed as much.

I would expect that every job would implicitly get a switch while the job is running. I wouldn't expect to see switches in pipeline definitions. Conceptually (and maybe in practise too), pipelines are evaluated outside of Eio. They can't perform async operations and don't get switches. However, the result of evaluating a pipeline may be that the engine needs to start some jobs running.

> I'm pretty sure the read and write functions for monitors are not quite right, I think there's a cleaner way to keep cancellation in the library compared to how I've done it in a few places.

Yeah, `Current.Switch` should get replaced with `Eio.Switch` eventually. But it's good to keep the code similar initially and clean it up later.

> There's no nice way it seems to support redirecting stderr to stdout in luv/libuv so I've used a socket pair... probably not the most efficient ^^"

That's annoying. Having separate sockets means the output can get out of sync between the two (e.g. app writes to stdout, then stderr, then stdout, but OCurrent reads the two combined stdout writes first, then the stderr one).

Anyway, I think this gives us some good places where Eio needs to add more support (process management, tmp dirs, etc). I wouldn't expect OCurrent to be converted any time soon, though. For an actual port, I'd start by using `Lwt_eio` for everything and then move over one bit at a time, but for driving the improvement of Eio this is great work :-)

patricoferris:
Agreed that it's definitely not the way to try and do everything at once ^^ (please close the draft PR if it's noisy, otherwise I'll leave it here for others to find)

shonfeder:
Given the amount of conflicts here and (I assume) changes to Eio, I don't think there's any chance we are going to land this. Please send word if you disagree.

PR #390, opened 2022-11-30
Debugging macos mdx test failures.

tmcgilchrist:
This PR works for me locally on macOS 12.6.1 but continues to fail on the macOS workers.
Could people using macOS please test it locally to see if this fix works for them?

Mdx fix: https://github.com/realworldocaml/mdx/pull/409


tmcgilchrist:
The underlying issue with mdx still occurs with forking off external processes. We didn't fix that, rather @maiste ported the tests to alcotest https://github.com/ocurrent/ocurrent/pull/405

PR #391, opened 2022-12-06
Misc. patches: don't use `List.length` for comparisons, use Unicode ellipsis

MisterDA:


benmandrew (approved):

plugins/github/api.ml
```diff
@@ -544,7 +544,7 @@ module Head_ref = Monitor(struct
           prefix
           name
           target {
-            ...on Commit {
```
TheLortex, line 547: In the GraphQL language `...` should be used instead of the UTF-8 ellipsis `…` .

plugins/github/api.ml
```diff
@@ -591,7 +591,7 @@ module Refs = Monitor(struct
           node {
             name
             target {
-              ...on Commit {
```
TheLortex, line 594: Same

PR #392, opened 2022-12-06
process: don't write to stdin without user-supplied data

MisterDA:
Spotted this, I've been wary of left open file descriptors lately.
I'm not super familiar with the object system, is the coercion valid? The objective is to not be able to call the `stdin` method outside of the `match` expression.
We can't use the type system to prevent access in `check_output` because `Lwt_process.process_full` is the only way to access `stdout` and `stderr` independently.

### DIFF
--- a/lib/process.ml
+++ b/lib/process.ml
@@ -141,31 +141,44 @@ let add_shutdown_hooks ~cancellable ~job ~cmd proc =
       )
   )
 
-let exec ?cwd ?(stdin="") ?(pp_cmd = pp_cmd) ?pp_error_command ~cancellable ~job cmd =
+let exec ?cwd ?stdin ?(pp_cmd = pp_cmd) ?pp_error_command ~cancellable ~job cmd =
   let cwd = Option.map Fpath.to_string cwd in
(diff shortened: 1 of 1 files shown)


talex5:
Why is this a problem? If there's no input then giving the process a pipe that gets closed immediately seems fine (if slightly less efficient than passing `/dev/null`).

MisterDA:
> If there's no input then giving the process a pipe that gets closed immediately seems fine

Does the pipe get closed immediately?

I saw some code using `Current.Cmd.Raw.docker ["run"; "-i"]` (so telling Docker to read from it's stdin) but not actually giving anything to Docker's stdin. While considering removing the `-i` flag I dug in and thought this patch would help, we don't retain the file descriptor if we don't really need it.

talex5:
> Does the pipe get closed immediately?

Yes, `send_to` closes it after sending the input data, which means immediately if there isn't any.

MisterDA:
> > Does the pipe get closed immediately?
> 
> Yes, `send_to` closes it after sending the input data, which means immediately if there isn't any.

You're right, I missed that. Thanks!

PR #393, opened 2022-12-09
Expose `Process.check_status`

MisterDA:
Useful as a general tool in the box, when processes are run outside of ocurrent or lwt, but we'd like to keep the error reporting identical, for consistency.


lib/process.mli
```diff
@@ -1,3 +1,8 @@
+val check_status :
+  ?pp_error_command:(Format.formatter -> unit) ->
+  string * string array ->
+  Unix.process_status -> (unit, [> `Msg of string ]) result
```
tmcgilchrist, line 4: I think this should be:
```suggestion
    Unix.process_status -> unit Current_term.S.or_error
```

MisterDA: good catch, thanks!

PR #399, opened 2022-12-20
Shorten the PR labels

kit-ty-kate:
Having the body of the PR in the label makes it very hard to work with the admin interface of ocurrent. For example:

scrn-2022-12-20-13-17-24

### DIFF
--- a/plugins/github/api.ml
+++ b/plugins/github/api.ml
@@ -319,8 +319,8 @@ module Commit_id = struct
       end
     | x -> x
 
-  let pp f { owner; repo; id; hash; committed_date; message } =
-    Fmt.pf f "%s/%s@ %a@ %s@ %s (%s)" owner repo pp_id id (Astring.String.with_range ~len:8 hash) committed_date message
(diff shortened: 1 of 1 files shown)


MisterDA (approved):

benmandrew:
Why has the timestamp been removed as well? Seems like useful information that doesn't take much space.

MisterDA:
We could also keep only the first line and truncate at 50 characters.

kit-ty-kate:
> Seems like useful information that doesn't take much space.

what information does it show that is useful when looking at this interface?

kit-ty-kate:
> We could also keep only the first line and truncate at 50 characters.

I don't really see the usecase for that information

tmcgilchrist:
I agree that having all the context from the commit message is too much, I've been meaning to remove it as well. 👍🏻

MisterDA:
I'm fine with the change and the infos that are retained. The complete merge message is definitely too much, the PR title is sufficient.
I think that the commit date gives more human-friendly info than the git sha, but I don't mind too much seeing it go. Let's merge.

PR #400, opened 2023-01-02
Expose refresh pipeline timer

### LINKED ISSUE #398: Automatic refresh is annoying on large graphs
When looking at the job graph, i have the following use-case:
1) lookup some error log on the userside website
2) open the corresponding subgraph (takes 30s already)
3) lookup some corresponding docker peek logs to rebuild (takes 20s)
4) lookup some more logs on the userside website
5) the page reloads and i have to wait 30s again...
6) leave it to reload, go back to it and it in a reload cycle once again

I don't personally need the automatic refresh feature added in https://github.com/ocurrent/ocurrent/pull/277 at all.
Is there a way to disable it?

MisterDA:
Applications can now set or disable this timer. Fixes #398. The timer is disabled by default, so I'll re-enable it in applications (the deployer and ocaml-ci-local).

### DIFF
--- a/lib_web/site.ml
+++ b/lib_web/site.ml
@@ -17,13 +17,14 @@ type t = {
   session_backend : Sess.backend;
   router : t raw Routes.router;
   nav_links : (string * string) list;   (* Label, path *)
+  refresh_pipeline : int option;
 }
 
 class type raw_resource = [t] raw
 
 let allow_all _ _ = true
 
-let v ?(name="OCurrent") ?authn ?(secure_cookies=false) ~has_role routes =
+let v ?(name="OCurrent") ?authn ?(secure_cookies=false) ?refresh_pipeline ~has_role routes =
   let db = Lazy.force Current.Db.v in
   let router = Routes.one_of routes in
   let nav_links = routes |> List.filter_map (fun route ->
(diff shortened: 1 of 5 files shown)


lib_web/site.ml
```diff
@@ -17,13 +17,14 @@ type t = {
...
   nav_links : (string * string) list;   (* Label, path *)
+  refresh_pipeline : int option;
 }
 
 class type raw_resource = [t] raw
 
 let allow_all _ _ = true
 
-let v ?(name="OCurrent") ?authn ?(secure_cookies=false) ~has_role routes =
+let v ?(name="OCurrent") ?authn ?(secure_cookies=false) ?refresh_pipeline ~has_role routes =
```
maiste, line 27: Do we want to keep the old behaviour? In case yes, shouldn't we set it as follow:
```suggestion
let v ?(name="OCurrent") ?authn ?(secure_cookies=false) ?(refresh_pipeline=60) ~has_role routes =
```

maiste (approved):
LGTM :+1:

lib_web/site.ml
MisterDA: We cannot disable the timer if we do it this way. I think `int option option` is too much just for this.

benmandrew (approved):
LGTM

PR #401, opened 2023-01-23
OCurrent SSH plugin

mtelvers:
This plugin creates OCurrent pipeline nodes which perform arbitrary SSH commands.

### DIFF
--- a/doc/examples/dune
+++ b/doc/examples/dune
@@ -10,7 +10,8 @@
                github_app
                rpc_server
                rpc_client
-               confirmation)
+               confirmation
+               example_ssh)
  (package current_examples)
  (libraries
    capnp-rpc
@@ -28,6 +29,7 @@
    current_gitlab
    current_rpc
    current_web
+   current_ssh
    dockerfile
    duration
    fmt.tty
--- a/doc/examples/dune
+++ b/doc/examples/dune
@@ -11,7 +11,7 @@
                rpc_server
                rpc_client
                confirmation
-               example_ssh)
+               ssh_command)
  (package current_examples)
  (libraries
    capnp-rpc
--- a/doc/examples/example_ssh.ml
+++ b/doc/examples/example_ssh.ml
@@ -1,0 +1,55 @@
+(* Usage: example_ssh.exe host args
+
+   Given a host and arguments, runs the command every 30 minutes.
+
+   e.g. Run `ls -phil` on host host.example.com
+   $ dune exec -- ./doc/examples/example_ssh.exe -- host.example.com tar -cvf bin.tar.gz /bin
+
+   Check the output and run history via http://localhost:8080
+*)
+
+let program_name = "current_ssh"
+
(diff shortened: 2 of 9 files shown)


doc/examples/example_ssh.ml
```diff
@@ -0,0 +1,55 @@
...
+let args =
+  Arg.value @@
+  Arg.pos_right 0 Arg.string ["ls"] @@
+  Arg.info
+    ~doc:"Command to run."
+    ~docv:"ARGS"
+    []
+
+let cmd =
+  let doc = "Build the head commit of a local Git repository using Docker." in
```
maiste, line 51: The doc here doesn't seem to be the right one.

plugins/ssh/current_ssh.ml
```diff
@@ -0,0 +1,8 @@
+open Current.Syntax
+
+module R = Current_cache.Output(Run)
+
+let run ~schedule ~key host args =
+  Current.component "ssh@,%s" host |>
+  let> args = args in
+  R.set ~schedule host key { Run.Value.args }
```
maiste, line 8: I might need clarification, but shouldn't the key be derived from the args? How is the key supposed to change?

maiste (commented):
Thanks for the PR! Few comments about styling and clarification.

doc/examples/dune
```diff
@@ -10,7 +10,8 @@
                github_app
                rpc_server
                rpc_client
-               confirmation)
+               confirmation
+               example_ssh)
```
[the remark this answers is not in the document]
mtelvers: Wouldn't that be confusing as there is already the command `ssh`?

doc/examples/example_ssh.ml
mtelvers: It looks like I started with another example and forgot to change that ;-)  I'll update it!

plugins/ssh/current_ssh.ml
mtelvers: Definitely not, because then you couldn't run the same command twice.  You pass the key when you invoke the call `Current_ssh.run ... ~key:"my-key" ... `.  You may want to run `ls` twice, perhaps once before and once after some other commands.  The first would be called with `~key:"ls1"` and the second as `~key:"ls2"`.

maiste: Thanks for the clarification! That's clearer. My concern is that the `key` word in this way seems to be tight to `OCurrent`. I would use something like `label` instead. However, this is opinionated, so feel free to ignore it :)

mtelvers: I went with `key` to match the definition in `lib_cache/s.ml`

doc/examples/dune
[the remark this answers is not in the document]
mtelvers: I've renamed it `ssh_command.ml`

maiste (approved):
Thanks for the changes, LGTM :+1:

PR #402, opened 2023-01-25
Github PRs: pulling labels

ElectreAAS:
This PR allows `pr_info` to hold label information.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,10 @@ Web UI:
 - Allow controlling and disabling the refresh timer for pipelines
   graphs and job page (@MisterDA, #398, #400)
 
+Plugins:
+
+- GitHub: Added information about labels in Api.Ref.pr_info (@ElectreAAS, #402)
+
 ### v0.6.3
 
 Web UI:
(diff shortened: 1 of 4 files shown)


maiste (commented):
Thanks for the PR! That's a nice to have. Could you please add an entry to the [CHANGES.md](https://github.com/ocurrent/ocurrent/blob/master/CHANGES.md) file?

ElectreAAS:
> Thanks for the PR! That's a nice to have. Could you please add an entry to the [CHANGES.md](https://github.com/ocurrent/ocurrent/blob/master/CHANGES.md) file?

Yes, done!

maiste:
> @maiste do you think that we should also provide the label color, along with its name ?

Not sure that it brings valuable information... However, if you think it can be useful in one of the ocurrent projects, no problem to add it. In this case, it should be an associative list with `(string * color) list`.

The use case I see is to tweak the behaviour in the function of the labels, not the colour. WDYT?

TheLortex (approved):

maiste:
Wait for the CI to goes green before merging :)

PR #403, opened 2023-01-26
Add Current_term.option_iter

kit-ty-kate:
Same as `Current_term.list_iter` but for the option type instead of list.


maiste (approved):

kit-ty-kate:
all green

Issue #384, opened 2022-11-08
Builds are failing due to timeout 


novemberkilo:
Jobs are timing out after the tests run - [example](https://ci.ocamllabs.io/github/ocurrent/ocurrent/commit/09741e/variant/alpine-3.15-4.14_opam-2.1)

Investigation shows that `dune runtest` does not exit cleanly:

```
❯ dune runtest
File "plugins/git/test/dune", line 1, characters 0-52:
1 | (mdx
2 |  (packages current_git)
3 |  (package current_git))
ocaml-mdx-test: internal error, uncaught exception:
                Unix.Unix_error(Unix.EINTR, "waitpid", "")
```

maiste:
Investigating. It seems to be related to mdx internal issue.

maiste:
We have two commits here to compare the working and broken behaviour:
* Woring: https://ci.ocamllabs.io/github/ocurrent/ocurrent/commit/cd9156e800915aac3149affaebd58518ee03f653/variant/alpine-3.15-4.14_opam-2.1
* Broken: https://ci.ocamllabs.io/github/ocurrent/ocurrent/commit/b75be989d67034e77e4a77316d342972d75ecc61/variant/alpine-3.15-4.14_opam-2.1

maiste:
After digging into the investigation, the issue seems unrelated to`mdx`. Indeed, if you look at the last commit that works and the first one that breaks, you can see the `mdx` version is precisely the same, and so are the other packages.

However, one point that seems to change is the git version in the docker image.  This is the patch that shows the difference between the package in the two ubove version:
```diff
 curl-7.80.0-r3 x86_64 {curl} (MIT) [installed]
< expat-2.4.9-r0 x86_64 {expat} (MIT) [installed]
---
> curl-7.80.0-r4 x86_64 {curl} (MIT) [installed]
> expat-2.5.0-r0 x86_64 {expat} (MIT) [installed]
19c19
< git-2.34.4-r0 x86_64 {git} (GPL-2.0-or-later) [installed]
---
> git-2.34.5-r0 x86_64 {git} (GPL-2.0-or-later) [installed]
29c29
< libcurl-7.80.0-r3 x86_64 {curl} (MIT) [installed]
---
> libcurl-7.80.0-r4 x86_64 {curl} (MIT) [installed]
41,42c41,42
< libx11-1.7.2-r0 x86_64 {libx11} (custom:XFREE86) [installed]
[...]

PR #405, opened 2023-02-06
Rewrite the current_git test to avoid `mdx` troubles.

maiste:
This PR deactivates `mdx` tests until we figure out why the tests are creating a timeout. They will be reactivated as soon as we have released the `0.6.4` version.

My first version tried to move to `mdx.0.3` stanza, but it didn't solve the issue. Then I tried only to disable the test on `MacOS`, but, unfortunately, I reached this bug with dune (https://github.com/ocaml/dune/issues/6938). The final version comments the tests.

### DIFF
--- a/plugins/git/test/git_test.ml
+++ b/plugins/git/test/git_test.ml
@@ -1,0 +1,193 @@
+open Lwt.Infix
+open Current.Syntax
+
+module Cmd = struct
+  let exec_or_fail ?cwd ~name cmd =
+    let cmd_s = String.concat " " cmd in
+    let cmd = Array.of_list cmd in
+    Lwt_process.exec ?cwd ("", cmd) >|= function
+    | Unix.WEXITED n ->
+        Alcotest.(check int) (Printf.sprintf "Process %s: %s" name cmd_s) 0 n
+    | Unix.WSTOPPED _ -> Alcotest.fail "Process stopped."
+    | Unix.WSIGNALED _ -> Alcotest.fail "Process received signal."
+
+  let mkdir ?cwd dir =
+    let cmd = [ "mkdir"; "-p"; dir ] in
+    exec_or_fail ?cwd ~name:"mkdir" cmd
+
+  let rm ?cwd cmd =
+    let cmd = "rm" :: cmd in
+    exec_or_fail ?cwd ~name:"rm" cmd
+
+  let mv ?cwd origin target =
+    let cmd = [ "mv"; origin; target ] in
+    exec_or_fail ?cwd ~name:"mv" cmd
+
+  let touch ?cwd file =
+    let cmd = [ "touch"; file ] in
+    exec_or_fail ?cwd ~name:"touch" cmd
+
+  let echo_to ?(cwd = "./") file content =
+    let file = Filename.concat cwd file in
+    Lwt_io.(
+      with_file ~mode:Output file (fun cout -> Lwt_io.write_line cout content))
+
+  let git ?cwd cmd =
+    let cmd = "git" :: cmd in
+    exec_or_fail ?cwd ~name:"git" cmd
+
+  let git_with ?cwd ~path cmd =
+    let cmd = "-C" :: path :: cmd in
+    git ?cwd cmd
+end
+
+let results, push_result = Lwt_stream.create ()
+
+module Show_files = struct
+  type t = unit
+
+  let id = "show-files"
+
+  module Key = struct
+    include Current_git.Commit
+
+    let digest t = Current_git.Commit_id.digest (id t)
+  end
+
+  module Value = Current.Unit
+
+  let build () job commit =
+    Current.Job.start job ~level:Current.Level.Harmless >>= fun () ->
+    Current_git.with_checkout ~job commit (fun tmpdir ->
+        let files =
+          Sys.readdir (Fpath.to_string tmpdir)
+          |> Array.to_list
+          |> List.filter (fun x -> x.[0] <> '.')
+          |> List.sort String.compare
+        in
+        push_result (Some files);
+        Lwt.return (Ok ()))
+
+  let pp = Current_git.Commit.pp
+  let auto_cancel = false
+end
+
+module SF = Current_cache.Make (Show_files)
+
+let show_files commit =
+  Current.component "show_files"
+  |> let> commit = commit in
+     SF.get () commit
+
+let init root =
+  let cwd = Fpath.to_string root in
+  let dir = "sub" in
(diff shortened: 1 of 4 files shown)


talex5:
> Then I tried only to disable the test on MacOS, but, unfortunately, I reached this bug with dune (https://github.com/ocaml/dune/issues/6938).

The bug there is that the tests are *built* even when disabled, which is a problem if the code only builds on one platform. But that shouldn't apply to MDX, I wouldn't have thought (please add to the bug report if MDX tests still run when disabled).

Disabling failing tests without understanding why they're failing seems like a bad idea. When did the problem start?

maiste:
> The bug there is that the tests are built even when disabled, which is a problem if the code only builds on one platform. But that shouldn't apply to MDX, I wouldn't have thought (please add to the bug report if MDX tests still run when disabled).

On the `OCaml-ci` MacOS worker, they seemed stuck in the `mdx` infinite loop we had in #384. It has the same behaviour as when there was no `enabled_if`. I don't have access to a `Mac` computer to test that locally, however...

> Disabling failing tests without understanding why they're failing seems like a bad idea. When did the problem start?

I agree. It was only to do the release and reactivate them afterwards. This issue is the same as the one mentioned in #384: `mdx` is stuck before timing out with the `git submodule` test.

From an offline discussion with @tmcgilchrist, I will try to rewrite the `mdx` test with `Alcotest` to ensure we keep the same coverage.

emillon:
FYI, I'm seeing timeouts on opam-repo-ci too. This is likely dependent on the version of the `git` binary used.

maiste:
Thanks, @emillon, for the follow-up.
My goal by rewriting it using `alcotest` is to be able to set a time lower than 1 hour. This way, even if it fails because of the `git` version, we won't have to wait this time to detect the error.

maiste:
The tests have been rewritten in `Alcotest`. This way, we won't suffer from any outside issues that could be caused by `mdx` . If we are stuck again in the `Timeout` issue, it will fail in __2 minutes__ instead of __1 hour__, making it easier to test.

This PR needs to be merged before being able to release `0.6.4`.

plugins/git/test/git_test.ml
```diff
@@ -0,0 +1,193 @@
...
+    let cmd_s = String.concat " " cmd in
+    let cmd = Array.of_list cmd in
+    Lwt_process.exec ?cwd ("", cmd) >|= function
+    | Unix.WEXITED n ->
+        Alcotest.(check int) (Printf.sprintf "Process %s: %s" name cmd_s) 0 n
+    | Unix.WSTOPPED _ -> Alcotest.fail "Process stopped."
+    | Unix.WSIGNALED _ -> Alcotest.fail "Process received signal."
+
+  let mkdir ?cwd dir =
+    let cmd = [ "mkdir"; "-p"; dir ] in
```
MisterDA, line 15:
```suggestion
    let cmd = [ "mkdir"; dir ] in
```

The original script doesn't use `-p`.

plugins/git/test/git_test.ml
```diff
@@ -0,0 +1,193 @@
...
+  Cmd.mkdir ~cwd dir >>= fun () ->
+  Cmd.git ~cwd [ "init"; "-q"; dir ] >>= fun () ->
+  let file = "sub/file" in
+  Cmd.echo_to ~cwd file "sub" >>= fun () ->
+  Cmd.git_with ~cwd ~path:"sub" [ "config"; "user.name"; "Name" ] >>= fun () ->
+  Cmd.git_with ~cwd ~path:"sub" [ "config"; "user.email"; "test@example.com" ]
+  >>= fun () ->
+  Cmd.git_with ~cwd ~path:"sub" [ "add"; "file" ] >>= fun () ->
+  Cmd.git_with ~cwd ~path:"sub"
+    [ "commit"; "-q"; "-a"; "-m"; "'Initial submodule commit'" ]
```
MisterDA, line 96:
```suggestion
    [ "commit"; "-q"; "-a"; "-m"; "Initial submodule commit" ]
```
Quotes were used for the shell, they should be removed.

plugins/git/test/git_test.ml
```diff
@@ -0,0 +1,193 @@
...
+  >>= fun () ->
+  Cmd.git_with ~cwd ~path:"main"
+    [ "commit"; "-q"; "-a"; "-m"; "Initial main commit" ]
+
+let remove root =
+  let cwd = Fpath.to_string root in
+  Cmd.rm ~cwd [ "main/.gitmodules" ] >>= fun () ->
+  Cmd.rm ~cwd [ "-r"; "main/sub" ] >>= fun () ->
+  Cmd.git_with ~cwd ~path:"main"
+    [ "commit"; "-q"; "-a"; "-m"; "'Remove submodule'" ]
```
MisterDA, line 117:
```suggestion
    [ "commit"; "-q"; "-a"; "-m"; "Remove submodule" ]
```
Same.

plugins/git/test/git_test.ml
```diff
@@ -0,0 +1,193 @@
...
+  let cwd = Fpath.to_string cwd in
+  Cmd.mv ~cwd "sub" "newsub" >>= fun () ->
+  Cmd.echo_to ~cwd "newsub/file2" "sub2" >>= fun () ->
+  Cmd.git_with ~cwd ~path:"newsub" [ "add"; "file2" ] >>= fun () ->
+  Cmd.git_with ~cwd ~path:"newsub" [ "config"; "user.name"; "Name" ]
+  >>= fun () ->
+  Cmd.git_with ~cwd ~path:"newsub"
+    [ "config"; "user.email"; "test@example.com" ]
+  >>= fun () ->
+  Cmd.git_with ~cwd ~path:"newsub" [ "commit"; "-q"; "-a"; "-m"; "'sub2" ]
```
MisterDA, line 137:
```suggestion
  Cmd.git_with ~cwd ~path:"newsub" [ "commit"; "-q"; "-a"; "-m"; "sub2" ]
```

MisterDA (changes requested):
Thank you Etienne, it's a faithful rewrite of the mdx code.

MisterDA (approved):

maiste:
The MacOS worker seems to have a struggle with finding `pkg-conf`.  Should we consider [this](https://ci.ocamllabs.io/github/ocurrent/ocurrent/commit/38c42b983238cdc0b9bc00ef7e889dc62095281c/variant/macos-homebrew-5.0.0_arm64_opam-2.1#L627-627) as an `ocurrent` issue or a MacOS worker one?

We can consider merging this PR if it's not a MacOS one.

PR #407, opened 2023-02-14
Remove explicit tls dependency

MisterDA:
`tls` is already an implicit dependency, and the `lower-bound` linter find 0.12.5 without the explicit dependency, so all seems fine.


talex5:
This is quite risky. The dependency is there to make sure `cohttp-lwt-unix` gets built with the right options (it uses conduit, which uses depopts).

MisterDA:
Ok, I was wondering if something like what you describe happened, but since it defaults to `tls` I thought the explicit dependency wasn't required. Closing for safety!

PR #410, opened 2023-02-20
Generate opam files using Dune

MisterDA:
The classic PR. This updates the build instructions of the packages and reorders the lists lexicographically.


tmcgilchrist:
Looks good to me.

Just check the version bounds on mirage-crypto and mdx are ok. There are
some PRs around to give them lower bounds. In particular mdx is buggy in
older versions.

PR #415, opened 2023-03-02
Bump constraint to OCaml 4.12

MisterDA:
current_incr already requires it, so forward the constraint here to make it clearer.


tmcgilchrist (approved):
Please add a CHANGES.md entry for this.

PR #417, opened 2023-03-21
Porting OCurrent to Eio

benmandrew:
Uses the experimental `capnp-rpc` Eio port in: https://github.com/mirage/capnp-rpc/pull/256


benmandrew:
Title changed due to Lwt being used in a lot more than just `current_rpc`.

PR #426, opened 2023-05-23
plugins/curl

mtelvers:
DIG/CURL plugin for https://github.com/ocurrent/ocurrent-observer to address https://github.com/ocaml/infrastructure/issues/42


plugins/curl/address.ml
```diff
@@ -0,0 +1,40 @@
...
+let pp f { fqdn = _; ip } = Fmt.pf f "%s" ip
+
+let pp_short f t =
+  Fmt.string f @@ Astring.String.with_range ~len:6 (hash t)
+
+let expand a =
+  Current.component "list" |>
+  let> a = a in
+  Current.Primitive.const (
+    String.split_on_char '\n' a.Dig.Value.result |>
```
maiste, line 26:
```suggestion
let expand dns  =
  Current.component "list" |>
  let> dns = dns in
  Current.Primitive.const (
    String.split_on_char '\n' dns.Dig.Value.result |>
```
Proposition to make it more readable. Feel free to ignore :+1:

maiste (approved):
2 suggestions in the naming convention. LGTM :+1:

avsm:
I don't quite see what curl and dig have to do with each other -- why not just have separate plugins for each, or generalise into a `Current_lookup` or `Current_health_check` or something that reflects the use here? I can imagine expanding to other protocols in the future such as a `netcat` to check if a TCP port is open, or a `ping`

plugins/curl/dig.ml
```diff
@@ -0,0 +1,47 @@
...
+  type t = {
+    fqdn : string;
+    result : string;
+  } [@@deriving yojson]
+    
+  let digest { fqdn; result } = 
+    Yojson.Safe.to_string @@ `Assoc [
+      "fqdn", `String fqdn;
+      "result", `String result;
+    ]
```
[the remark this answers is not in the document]
mtelvers: Thanks.  Updated.

plugins/curl/address.ml
mtelvers: I went with `a` for address but happy to change it to `dns`

mtelvers:
@avsm Yes, I agree.  I have renamed it `ocurrent_health_check`, added a ping module and updated http://observer.ocamllabs.io

plugins/health_check/ping.ml
```diff
@@ -0,0 +1,20 @@
...
+
+module Key = Address
+
+module Value = Current.Unit
+
+module Outcome = Current.Unit
+
+let publish No_context job key _ =
+  Current.Job.start job ~level:Current.Level.Dangerous >>= fun () ->
+  Current.Process.exec ~cancellable:true ~job ("", [| "ping"; "-c"; "4"; key.Key.ip|])
```
avsm, line 15: One annoying unportable thing is that this'll only work on `ping` commands that support both IPv4 and IPv6.  On macOS, for example, you have to use `ping6` for IPv6 endpoints.

mtelvers: I have updated the code to use `ping6` for IPv6 addresses.  On the Linux machines I checked: Ubuntu, Debian, and Raspbian, this was a symlink to `ping`.

plugins/health_check/ping.ml
```diff
@@ -0,0 +1,23 @@
...
+
+module Value = Current.Unit
+
+module Outcome = Current.Unit
+
+let publish No_context job key _ =
+  Current.Job.start job ~level:Current.Level.Dangerous >>= fun () ->
+  let cmd = match String.contains key.Key.ip ':' with
+    | true -> "ping6"
+    | false -> "ping" in
```
MisterDA, line 17:
```suggestion
  let cmd = if String.contains key.Key.ip ':' then "ping6" else "ping" in
```

plugins/health_check/dig.ml
```diff
@@ -0,0 +1,43 @@
...
+    match Yojson.Safe.from_string s |> of_yojson with
+    | Ok x -> x
+    | Error e -> failwith e
+end
+
+let id = "dig"
+
+let build No_context job { Key.fqdn } =
+  Current.Job.start job ~level:Current.Level.Mostly_harmless >>= fun () ->
+  let cmd = ("dig", [| "dig"; "@8.8.8.8"; "+noall"; "+answer"; fqdn; "A"; fqdn; "AAAA" |]) in
```
MisterDA, line 36: Is it possible to add multiple DNS servers? I'd love it if you tried an IPv6 server first, such as `2001:4860:4860::8888` (also from Google).

MisterDA:
Maybe we could use [happy-eyeballs](https://github.com/roburio/happy-eyeballs) instead of dig? I'm concerned that dig may not be available on the platform (Windows…). Happy-eyeballs also gracefully handles IPv6 to IPv4 degradation, and it's OCaml all the way down.

PR #433, opened 2023-08-10
Allow users to get serialised key of job

patricoferris:
For developing different UIs for current_web this is quite useful to expose


shonfeder (approved):
LGTM! Thanks for the improvement and for your patience :)

PR #439, opened 2023-10-27
Current_web: fix access to logfiles with spaces in the name

edwintorok:
`Uri.path` is documented to return an encoded path, so we need to decode it before using it. Otherwise some job logs are inaccessible from the web UI:
```
Job log /home/edvint/git/personal_pipelines2/var/job/2023-10-27/110201-bump%20commit-272052.log does not exist
```


tmcgilchrist:
Thanks @edwintorok

Issue #435, opened 2023-09-07
Rendering pipeline.svg on ocaml-docs-ci results in Stack overflow


tmcgilchrist:
Requesting the [pipeline.svg](https://docs.ci.ocaml.org/?compile=&) on OCaml Docs CI results in a Stack Overflow.

benmandrew:
The stack overflow only occurs when rendering graphs that are too big to be useful anyway. Collapsible groups and other restructuring can be used to make the graph more manageable.

PR #441, opened 2023-11-07
Tail-recursive graph rendering to prevent stack overflows

benmandrew:
The function `pp_dot` is causing stack overflows for very large graphs, see: https://github.com/ocurrent/ocurrent/issues/435. The function writes OCurrent graphs into DOT format (which are later translated to SVG for rendering in the web UI).

I have attempted to rewrite the function in CPS (continuation passing style) form to make it tail-recursive, but have not yet succeeded in getting the tests to pass. The function is large with complex control flow.

It may be preferred to write the function in an iterative form, using an explicit stack instead of the implicit one given by recursive functions. The explicit stack would be stored in the heap, which is less likely to overflow.

### DIFF
--- a/lib_term/analysis.ml
+++ b/lib_term/analysis.ml
@@ -18,7 +18,7 @@ module Make (Meta : sig type t end) = struct
         match t.ty with
         | Constant None ->
           begin match t.bind with
-            | Some ctx -> aux f ctx
+            | Some ctx -> (aux [@tailcall]) f ctx
             | None ->
               match Current_incr.observe t.v with
               | Error (_, `Active _) -> Fmt.string f "(input)"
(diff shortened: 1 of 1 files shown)


benmandrew:
All recursive call sites in the auxiliarly function `aux` in `pp_dot` have `[@tailcall]` compiler annotations. However, I'm not able to generate large enough OCurrent graphs locally to cause the original stack overflow, so I can't see for myself whether the optimisation is being done by the compiler.

It would be useful to test this PR live on OCaml Docs CI.

benmandrew:
The `[@tailcall]` annotation appears to not be supported in this manner below OCaml 4.13. It's up for discussion whether this lower-bound version can be bumped.

talex5:
I'm not sure it's useful to generate a graph so large it would overflow the OCaml stack. graphviz is unlikely to be able to render it sensibly. It might be better to add some groups to the pipeline so they then can be expanded and collapsed as needed.

If you really need it though, you can always increase the stack limit (e.g. `OCAMLRUNPARAM=l=1000000000`).

benmandrew:
Closing for the reasons detailed above.

PR #442, opened 2023-11-13
Added option to directly construct Current_git.Commit.t values

ElectreAAS:
I added the `Current_git.v` function, and I am wondering why it didn't exist in the first place.
Anyway I think there's no way around doing it that way for my use case (see https://github.com/ocurrent/current-bench/pull/482).


benmandrew (approved):

ElectreAAS:
Thanks!

PR #443, opened 2024-02-05
Add note about dependancies required before trying to run examples.

mdales:
Checking out OCurrent and trying to run the example in the documents assumes you have done some things between those two steps that I didn't find documented, and it took me a while to figure out. This PR just makes those steps explicit, hopefully saving others in the future some time.

### DIFF
--- a/doc/index.md
+++ b/doc/index.md
@@ -1,5 +1,15 @@
 Welcome to the OCurrent documentation!
 
+### Prerequisites
+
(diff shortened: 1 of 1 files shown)


benmandrew:
Hi @mdales, I ran into the same issue and realised the submodule can be removed (done in #444), so this PR is no longer necessary. But thank you for contributing!

PR #447, opened 2024-07-16
Fix GitHub CheckRun InProgress status identifier

zazedd:
`inprogress` is not valid. `in_progress` is the correct identifier as per [the docs](https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28)


shonfeder (approved):
Thank you for the fix!

PR #451, opened 2024-08-22
Add retry utility

### LINKED ISSUE #450: Expose a combinator for retry logic
This mechanism will help us deal with various operations that can be flakey. In general, this will be useful to have for anything making network calls, but it will at least let us address needs like the following in a general way:

- https://github.com/ocurrent/ocurrent/issues/199#issue-621012391
- https://github.com/ocurrent/docker-base-images/issues/211
- https://github.com/ocurrent/ocaml-docs-ci/pull/135
- https://github.com/ocurrent/ocaml-docs-ci/pull/105

We considered several approaches to this, including 

- A combinator at the level `Current_term` that would make a whole current retryable -- This is almost always going to be at too high a level of granularity: we need retries in the level of `Lwt` operations.
- Adding a label to the `Process` functions to trigger retry logic -- This is overly specialized, as it is easy to imagine cases where operations that do not go through subprocesses would like to make use of retry logic (e.g., network requests using `Cohttp`).

I think the sweet spot here is to adopt an adapted version of https://github.com/ocurrent/ocaml-docs-ci/blob/main/src/lib/retry.mli into ocurrent. This can then exposed in the public api, and used to address key needs within ocurrent plugins, allowing us to solve, e.g., https://github.com/ocurrent/docker-base-images/issues/211 by adding a sensible retry to https://github.com/ocurrent/ocurrent/blob/d57b358919bb2857548679488a1ce8fc96df0d51/plugins/docker/push_manifest.ml#L61.

shonfeder:
Closes #450
Contributes to https://github.com/ocurrent/docker-base-images/issues/211

I initially looked at adapting https://github.com/ocurrent/ocaml-docs-ci/blob/main/src/lib/retry.mli, but realized that the interface for this is not very reusable or well formed (IMO): e.g., using an empty list to signal a success is not very nice.

The approach suggested here provides three small, composable combinator to deal with retry logic in a flexible way. The base function, `Retry.on_error` will produce a (potentially) unlimited sequence of attempts, and the particular functionality a user may want (like adding sleeps, or erroring after a set number of retries) is provided by short functions working over the stream.

### DIFF
--- a/lib/current.ml
+++ b/lib/current.ml
@@ -333,6 +333,7 @@ module Process = Process
 module Switch = Switch
 module Pool = Pool
 module Log_matcher = Log_matcher
+module Retry = Retry
 
 module Job = struct
   include Job
--- a/lib/current.mli
+++ b/lib/current.mli
@@ -460,3 +460,5 @@ module Log_matcher : sig
   val drop_all : unit -> unit
   val analyse_string : ?job:Job.t -> string -> string option
 end
+
+module Retry = Retry
--- a/lib/retry.ml
+++ b/lib/retry.ml
@@ -1,0 +1,72 @@
+open Lwt.Syntax
+
+let default_sleep_duration n' =
+  let base_sleep_time = 2.0 in
+  let n = Int.to_float n' in
+  n *. base_sleep_time *. Float.pow 1.5 n
+
+type ('retry, 'fatal) error =
+  [ `Retry of 'retry
+  | `Fatal of 'fatal
+  ]
+
+let pp_error ?retry ?fatal fmt err =
+  let default fmt _ = Fmt.pf fmt "<opaque>" in
+  let pp_retry = Option.value retry ~default in
+  let pp_fatal = Option.value fatal ~default in
+  match err with
+  | `Retry r -> Fmt.pf fmt "retryable error '%a'" pp_retry r
+  | `Fatal f -> Fmt.pf fmt "fatal error '%a'" pp_fatal f
+
+let equal_error ~retry ~fatal a b =
+  match a, b with
+  | `Retry a', `Retry b' -> retry a' b'
+  | `Fatal a', `Fatal b' -> fatal a' b'
+  | _ -> false
+
+type ('ok, 'retry, 'fatal) attempt = ('ok, ('retry, 'fatal) error) result
+
+let is_retryable = function
+  | Error (`Retry _) -> true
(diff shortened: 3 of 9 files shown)


mtelvers:
I haven't been able to test this PR despite extensive efforts, which I will continue tomorrow.  Testing this PR using a patched version of `ocurrent-deployer` doesn't help the root problem.  The retry code never runs because the failures we see occur in the `docker push`, which happens within `obuilder` running on the individual workers.  The push of the manifest, which `ocurrent-deployer` runs, isn't the part which fails.  The manifest operation binds all the individual images together into a multiarchitecture image.  This push is just a small json file.  This has been a eureka moment of where the problem lies.  Note, that while we only see `Docker.push_manifest` in `ocurrent-deployer`, we are ignoring the `OCluster.Raw.build_and_push` steps which cause the worker to push the images.  I think we need to add the retry logic to `plugins/docker/push.ml` instead of `plugins/docker/push_manifest.ml` (or perhaps as well as).

shonfeder:
>  I think we need to add the retry logic to plugins/docker/push.ml instead of plugins/docker/push_manifest.ml (or perhaps as well as).

Ah, that is very helpful. Sorry I missed this in previous diagnostics, but glad your testing was able to unearth it. Thanks you also for undertaking integration testing! I didn't intend to leave that to you but appreciate you digging in and correcting the target here. I'll add the change to `push` as well.

mtelvers:
`ocluster-worker` uses `lwt` only, not `ocurrent`; thus my suggestion of patching `plugins/docker/push.ml` is erroneous.

shonfeder:
I see. This clarifies the issue considerably, and I realize this is miss targeted a bit. The place we want to introduce retries first and foremost appears to be https://github.com/ocurrent/ocluster/blob/8fd1ca1da935154e6c532c45a4d8d4c6a2f9d307/worker/cluster_worker.ml#L98-L103 (judging from the logging noted in https://github.com/ocurrent/docker-base-images/issues/211#issuecomment-2221716687).

Fortunately the code proposed here is easily relocatable. I'll port this to ocluster for now, and we can get this problem resolved. Meanwhile we can think about where would be a good place fort his kind of utility to live to make it possible to share more widely, if the need arises.
Ocluster has [its own `process` module](https://github.com/ocurrent/ocluster/tree/8fd1ca1da935154e6c532c45a4d8d4c6a2f9d307/worker/process.ml).

PR #452, opened 2024-08-27
"docker login" now accepts user@host

mtelvers:
This PR allows a private registry to be used rather than always pushing to Docker Hub.

### DIFF
--- a/plugins/docker/cmd.ml
+++ b/plugins/docker/cmd.ml
@@ -13,7 +13,7 @@ let compose ~docker_context args =
   "", Array.of_list ("docker-compose" :: context_args docker_context @ args)
 
 let login ?config ~docker_context user =
-  docker ?config ~docker_context ["login"; "--password-stdin"; "--username"; user]
+  docker ?config ~docker_context (["login"; "--password-stdin"; "--username"] @ String.split_on_char '@' user)
 
 let pp f (prog, args) =
   if prog <> "" then Fmt.pf f "[%S] " prog;
--- a/plugins/docker/s.ml
+++ b/plugins/docker/s.ml
@@ -28,7 +28,8 @@ module type DOCKER = sig
       @param arch Select a specific architecture from a multi-arch manifest.
       @param schedule Controls how often we check for updates. If the schedule
                       has no [valid_for] limit then we will only ever pull once.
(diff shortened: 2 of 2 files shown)


plugins/docker/cmd.ml
```diff
@@ -13,7 +13,7 @@ let compose ~docker_context args =
   "", Array.of_list ("docker-compose" :: context_args docker_context @ args)
 
 let login ?config ~docker_context user =
-  docker ?config ~docker_context ["login"; "--password-stdin"; "--username"; user]
+  docker ?config ~docker_context (["login"; "--password-stdin"; "--username"] @ String.split_on_char '@' user)
```
shonfeder, line 16: Why expose this as `user@host` rather than as a new optional paramter `?host`?

My question here is motivated by the principle that it is generally preferable to to work on fully parsed, structured data, rather than encoding compound data into strings.

`pull` and `push` could also be extended with a `?host` argument, in a change that should be none breaking for almost all use cases.

WDYT?

shonfeder (approved):
Talked with mark sync. IMO, the API for this lib would be better with a labeled arg but it's not at all critical.

plugins/docker/cmd.ml
mtelvers: This is rather more tricky as `pull` and `push` need to conform to `Current_cache` which means they only accept a single _context_ parameter.  Thus, within the plugin, I have replaced the `string * string` type with `type auth = { user : string;  password : string;  server : string option; }`, but the API is still the user/password pair with the additional `?server` parameter.

plugins/docker/s.ml
```diff
@@ -28,7 +28,8 @@ module type DOCKER = sig
       @param arch Select a specific architecture from a multi-arch manifest.
       @param schedule Controls how often we check for updates. If the schedule
                       has no [valid_for] limit then we will only ever pull once.
-      @param auth If given do a "docker login" using this username/password pair before pulling. *)
+      @param auth If given, do a "docker login" using this username/password pair before pulling.
+                  If username is in the form "user@host", the registry at "host" will be used *)
```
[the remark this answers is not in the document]
mtelvers: Renamed: host to sever

PR #453, opened 2024-09-02
Add query entries for Github PR's: fork repository name and fork branch name

zazedd:
This PR adds two new fields to the pull request GraphQL request and their respective functions:
- `Commit.pr_fork_owner_name`: returns the name as `owner/name` of the fork from which the PR originated from, or `None` if it is a branch. Uses the field `headRepository.nameWithOwner` in the GraphQL request;
- `Commit.pr_fork_branch_name`: returns the branch name of that fork, or `None` if it is a regular (non-fork) branch. Uses the field `headRefName` in the GraphQL request;

This might be useful for pushing results (from the pipeline) to the fork where the PR originated.

---

A use case example can be found [here](https://github.com/zazedd/radiocarbon-pipeline/blob/1141cda9602d82962f65066752c3141f928b88ac/src/git_cache.ml#L243C5-L243C22), where I have vendored a version of OCurrent with these functions available.
The outputs are pushed to the fork repository of the person who submitted the pull request so that they can confirm whether they look correct or not.

### DIFF
--- a/plugins/github/api.ml
+++ b/plugins/github/api.ml
@@ -266,6 +266,8 @@ module Ref = struct
     title: string;
     labels: string list;
     bodyHTML: string;
+    branch_name: string;
+    fork: string;
   } [@@deriving to_yojson]
 
   type t = [ `Ref of string | `PR of pr_info ] [@@deriving to_yojson]
@@ -341,6 +343,16 @@ module Commit_id = struct
         Some (Astring.String.concat ~sep:"/" branch)
       | _ -> None)
     | `PR _ -> None
+    
+  let pr_fork_branch_name t =
+    match t.id with
+    | `Ref _ -> None
+    | `PR pr -> Some pr.branch_name
+
+  let pr_fork_with_owner t =
+    match t.id with
+    | `Ref _ -> None
+    | `PR pr -> Some pr.fork
 end
 
 module Repo_key = struct
@@ -610,6 +622,10 @@ module Refs = Monitor(struct
             headRefOid
             baseRefName
             title
+            headRefName
+            headRepository {
+              nameWithOwner
+            }
             labels(first: 100) {
               nodes {
                 name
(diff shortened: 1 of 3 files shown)


punchagan:
Thanks for the contribution @zazedd ! We'll take a look at it soon!

plugins/github/api.ml
```diff
@@ -341,6 +343,16 @@ module Commit_id = struct
         Some (Astring.String.concat ~sep:"/" branch)
       | _ -> None)
     | `PR _ -> None
+    
+  let pr_fork_branch_name t =
```
punchagan, line 347: What do you think about combining this functionality into the `branch_name` function above? Currently, it returns None for `PR` type.

plugins/github/api.ml
```diff
@@ -341,6 +343,16 @@ module Commit_id = struct
...
+    match t.id with
+    | `Ref _ -> None
+    | `PR pr -> Some pr.headRefName
+
+  let pr_fork_owner_name t =
```
punchagan, line 352: I find the `owner_name` name confusing sometimes. We use `_name` to represent name of something (for instance, `branch_name`, `pr_name`, etc.), and in this function we are trying to use it for `owner/name` representation. How about using something simpler like `pr_fork`?

plugins/github/api.ml
```diff
@@ -646,13 +662,30 @@ module Refs = Monitor(struct
     let hash = node / "headRefOid" |> to_string in
     let pr = node / "number" |> to_int in
     let title = node / "title" |> to_string in
+    let headRefName = node / "headRefName" |> to_string in
```
punchagan, line 665: I see that the names `headRefName` and `headNameWithOwner` come from the GitHub API, but would it be nicer to use something like `branch_name` and `fork`? It could be clearer to map to the names from the API in some cases, but I'm leaning towards the simpler names.

punchagan (commented):
Thanks for your contribution, @zazedd ! The functionality works nicely. 

I had a few minor "naming" related suggestions for your consideration. Feel free to disregard any of them, if they don't seem appropriate.

plugins/github/api.ml
shonfeder: Another option could be `pr_fork_with_owner`?

plugins/github/api.ml
shonfeder: tl;dr: I'm not sure the naming here matters too much... but I am mostly not sure :)

The suggestions here seem to fit with the naming of the accessor functions. And I can see value in keeping naming between the fields and accessor functions aligned. On the other hand, naming the field after the gql key follows the precedent of `bodyHTML`.

But then we don't follow the same practice here with `base` above. This leads me to think that we mabye don't have a consistent approach here between using the record as a more or less raw representation of the gql data vs. using it as a more refined/abstracted view.

plugins/github/api.ml
shonfeder: This would make sense to iff we can can always work with a PR fork head the same as way as we can with a branch in a `Ref`, otherwise we may not want to conflate them. I'm not sure off the top of my head whether they are interchangeable in all cases...

shonfeder (approved):

shonfeder:
Thanks indeed for the contribution, @zazedd!

plugins/github/api.ml
zazedd: Sorry for the delay in answering!

The branch names should be interchangeable, as long as we disambiguate a PR from a normal Ref and program accordingly.
It will break people's code where they expect this function to return `None` if the ref is a PR, though.

plugins/github/api.ml
zazedd: I like `pr_fork_with_owner`! I agree that `owner_name` is not a very good naming convention.

plugins/github/api.ml
punchagan: I agree, we probably don't have a consistent approach. But, it may be nicer to have `head` when we have a `base`.

plugins/github/api.ml
punchagan: Yes, I think so too.

plugins/github/api.ml
zazedd: I changed the variables according to @punchagan, I also think they make sense (especially given that the names GitHub gave them are more obscure)

shonfeder:
Thanks very much! I'm happy with this. I'll leave it to @punchagan to merge when happy with it!

PR #455, opened 2024-09-10
Raise and reraise exceptions with Stdlib rather than Lwt

MisterDA:
Lwt's documentation reads:

> In most cases, it is better to use `failwith s` from the standard
> library.

and

> Whenever possible, it is recommended to use `raise exn` instead, as
> raise captures a backtrace, while `Lwt.fail` does not. If you call
> `raise exn` in a callback that is expected by Lwt to return a
> promise, Lwt will automatically wrap `exn` in a rejected promise,
> but the backtrace will have been recorded by the OCaml runtime.
>
> For example, `bind`'s second argument is a callback which returns a
> promise. And so it is recommended to use `raise` in the body of that
> callback.
>
> Use `Lwt.fail` only when you specifically want to create a rejected
> promise, to pass to another function, or store in a data structure.

Prefer to capture backtraces to improve debugability.

Also use `Fmt.error` and `Fmt.error_msg`.


shonfeder:
I've struggled with that too a bit already :)

shonfeder (approved):
Looks great to me, and all cases fixed when I checked.

Thanks!

PR #456, opened 2024-10-03
doc(Current_web.Role.t): make comments visible in docs

edwintorok:
Need to be doc-comments to be visible in the generated documentation.


shonfeder (approved):
Thank you for the fix!

PR #457, opened 2024-10-08
Current_github: Allow `Current_git.Commit_id` to be created with an SSH URL

zazedd:
This PR changes the `Current_github.Commit_id.to_git` function to create a `Current_git.Commit_id.t` with either an HTTPS or SSH repo URL, and introduces an optional argument to the `Current_github.Commit.id` function to choose between the two.

---

This is useful if, for example, the remote repository is private and ssh is needed for authentication while fetching.

### DIFF
--- a/plugins/github/api.ml
+++ b/plugins/github/api.ml
@@ -301,8 +301,9 @@ module Commit_id = struct
     message : string;
   } [@@deriving to_yojson]
 
-  let to_git { owner; repo; id; hash; committed_date = _; message = _ } =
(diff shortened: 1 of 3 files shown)


shonfeder (approved):
This make sense to me and the patch looks clean, and shouldn't break users in general, thanks to use of an optional argument.

I'll leave this to settle for a few days in case any one else wants to review tho.

shonfeder:
Thanks again for the contribution!

PR #460, opened 2025-01-07
Support passing a specific interface to listen to for the website

mattam82:
This should allow the website to not listen to all interfaces by default (required for added security)

### DIFF
--- a/lib_web/current_web.ml
+++ b/lib_web/current_web.ml
@@ -70,17 +70,30 @@ let handle_request ~site _conn request body =
     | (`HEAD | `PUT | `OPTIONS | `CONNECT | `TRACE | `DELETE | `PATCH | `Other _) ->
       Utils.Server.respond_error ~status:`Bad_request ~body:"Bad method" ()
 
-let pp_mode f mode =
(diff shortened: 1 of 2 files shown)


mattam82:
Really not sure about my sexp hack here :)

mtelvers (changes requested):
Happy New Year. Thank you for the PR. This is a good addition, particularly useful in a test setup where authentication has yet to be implemented.

Rather than using a tuple, I would prefer a type as in `ocurrent/lib/config.ml`.  For example, we could have:

```
type t = {
  host : string option;
  port : Conduit_lwt_unix.server;
}
```

`current_web.mli` would then be

```
val run : ?mode:t -> Site.t -> ('a, [`Msg of string]) result Lwt.t
val cmdliner : t Cmdliner.Term.t
```

mtelvers (approved):
Great.  Thanks for the updates and the PR.

mattam82:
A release would be great if possible :) FTR, we use ocurrent as part of deploy.rocq-prover.org where we needed these features.

PR #472, opened 2026-05-28
docker: add ?os to Peek and Pull (default linux)

mtelvers:
Fixes `docker manifest inspect` peek for Windows images. The previous code hardcoded `os = "linux"` in the manifest filter, so peeking e.g. `ocaml/opam:windows-server-mingw-ltsc2025-ocaml-5.4` failed with `Failed to find arch "amd64" in manifest (Not_found)`. `os` defaults to `"linux"`, so existing callers are unaffected.

### DIFF
--- a/plugins/docker/current_docker.ml
+++ b/plugins/docker/current_docker.ml
@@ -134,8 +134,7 @@ module Make (Host : S.HOST) = struct
 
   let peek ?label ~arch ?(os="linux") ~schedule tag =
     let label = Option.value label ~default:tag in
-    let os_label = if os = "linux" then "" else "@," ^ os in
-    Current.component "peek %s@,%s%s" label arch os_label |>
+    Current.component "peek %s@,%s@,%s" label arch os |>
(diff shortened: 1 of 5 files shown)


plugins/docker/current_docker.ml
```diff
@@ -132,11 +132,12 @@ module Make (Host : S.HOST) = struct
     let> () = Current.return () in
     Raw.pull ~docker_context ~schedule ?arch ?auth ?server tag
 
-  let peek ?label ~arch ~schedule tag =
+  let peek ?label ~arch ?(os="linux") ~schedule tag =
     let label = Option.value label ~default:tag in
-    Current.component "peek %s@,%s" label arch |>
+    let os_label = if os = "linux" then "" else "@," ^ os in
```
talex5: I think that's just going to add `@,` as a literal separator. You might as well just print the OS even if it is Linux.

mtelvers: Thanks!

